PR #11, opened 2020-07-01
Added signup endpoint to server.

AneeshJalan:
Added signup endpoint to server.

Created an Auth controller and router that exposes and handles the /signup endpoint.
Also created a file containing some constant strings for use in sending back error
response messages.


AneeshJalan:
Yeah. The field names definitely need to be synced up. I can make the changes on the server side.

As for what the whitelisted document has, it doesn't matter what fields the document has. The most important information is the email. Everything else gets updated with the information passed in via the request body. One thing to note is that if the document has extra fields (other than email, first_name, last_name, major and graduation_year) then they will remain in the document even after the update.

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+  // Grab form information from request body or if body is undefined,
+  // parse from {}. (Helps avoid TypeError)
+  const { email,
+    password,
+    first_name,
+    last_name,
+    major,
+    graduation_year } = req.body || {};
+
+  if (!validInput(email, password, first_name, last_name, major, graduation_year)) {
```
godwinpang, line 38: validInput should probably just take req.body so you dont have to type in all these params lol

tbh if we assume (and this is a bad assumption) that the only user of this API is our frontend then we don't have to worry about request validation for now

an easier way to solve this would probably be to onboard something like [express-openapi-validator](https://github.com/cdimascio/express-openapi-validator#example-express-api-server)

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+    // Invalid user information
+    res.status(400).json({
```
godwinpang, line 40: a more idiomatic way of error handling in express is to pass in the error to next(), which then invokes express's internal error handler. this also means less res.status() everywhere

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+      msg: ERR_MSG.GENERIC_400_MSG,
+    });
+    return next();
+  }
+
+  const userDocID = await getUserDocFromEmail(email);
```
godwinpang, line 46: getUserDocFromEmail should throw error instead of returning undefined(?) this means we can leverage next() for error handling

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  if (!userDocID) {
+    // email is not whitelisted.
+    res.status(403).json({
+      msg: ERR_MSG.USER_NOT_WHITELISTED,
+    });
+    return next();
+  }
+
+  const newUser: admin.auth.UserRecord = await admin.auth().createUser({
```
godwinpang, line 56: more idiomatic to not mix promises with await/async syntax, can probably just wrap this await in a try catch to catch the error from createUser

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+    uid: userDocID,
+    email: email,
+    password: password
+  })
+    .then(userRec => {
```
godwinpang, line 61: essentially empty promise

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+      return userRec;
+    })
+    .catch(err => {
+      if (typeof (err.code.includes) === 'function' && err.code.includes('exists')) {
```
godwinpang, line 65: not a great way of validating error codes since includes might cause problems down the line; if firestore auth has error codes as types I'd rather we explicitly list which codes we're catching

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+        });
+        return undefined;
+      }
+    });
+
+  if (!newUser) {
+    return next();
+  }
+
+  await updateUserDoc(userDocID, {
```
godwinpang, line 84: same comment as above with mixing async/await with promises

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+      res.status(200).json({
+        userRecord: newUser,
+      });
+      return result;
+    })
+    .catch(err => {
+      // Could not update the user document with provided data.
+      // Delete account because we want to have this data in the user account at all times.
+      admin.auth().deleteUser(userDocID)
+        .then(res => {
```
godwinpang, line 101: nested promises are quite hard to read (although I blame js 2-space-indent for that) might be easier to do try catch instead

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+
+          // BUG TODO - This is a consequence of having an auth onDelete firebase cloud function
```
godwinpang, line 103: we can just delete the onDelete firebase cloud function if it's causing you troubles - those things are on their way out anyways

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+): boolean {
+  // Add form input validation here.
+  if (!email || !password || !first_name || !last_name || !major || !grad_year) {
+    return false;
+  }
+
+  return true;
+}
+
+async function getUserDocFromEmail(email: string): Promise<string> {
```
godwinpang, line 140: this function doesn't return a userdoc? it just returns the id of the doc?

src/routers/auth.router.ts
```diff
@@ -0,0 +1,12 @@
+import express from 'express';
+import * as controller from '../controllers/auth.controller';
+
+export const AuthRouter = express.Router();
+
+/**
```
godwinpang, line 6: we can come back to this later, but swagger has a good amount more we can use to document (error codes etc)

src/constants/ErrResponses.ts
```diff
@@ -0,0 +1,5 @@
+export const GENERIC_400_MSG = "The request object received had invalid syntax or missing information.";
+export const USER_NOT_WHITELISTED = "Cannot create an account since the email id provided does not have any affiliation with HKN. Visit hkn.ucsd.edu to know more.";
+export const USER_ALREADY_EXISTS = "The provided email is aready linked to an account. Please check the email you have entered.";
+export const FIREBASE_AUTH_CREATE_USER_ERR = "There was an internal error creating an auth account. Please try again later.";
+export const FIRESTORE_UPDATE_DOC_ERR = "There was an error modifying the user document. Please try again later."
```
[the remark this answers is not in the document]
AneeshJalan: Okay. I guess the last 2 are internal errors that probably need not be mentioned to the user. For the rest I think it is fine since we will eventually display some error message or prompt to the user when signup fails right?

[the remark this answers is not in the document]
godwinpang: :+1:

README.md
```diff
@@ -5,11 +5,11 @@ This is where we can run APIs.
 
 ## Running Server
 
-Type `heroku local` in your terminal (you may need to download Heroku)
+Type `npm run start` in your terminal.
```
godwinpang, line 8: thanks for updating readme

src/controllers/auth.controller.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  try {
+    await updateUserDoc('users', userDocID, {
+      first_name: firstname,
+      last_name: lastname,
+      major: major,
+      graduation_year: gradYear,
+    });
+
+    // Sign in to user account, send verification email and sign out.
```
godwinpang, line 109: better to wrap these 3 lines in a single function to maintain abstraction for ease of change later when we migrate to email service

godwinpang:
there's some stuff to cleanup (mainly the console.logs + abstractions) but go ahead and merge it in for now so thai's unblocked

godwinpang:
@thaigillespie looks like we didn't camel firstname/lastname on the frontend
This isn't a big deal (although we should fix it soon yay another small issue) but it highlights the need for a **design-first methodology** where we specify the public interface of the API ***before*** we start any code

Swagger/OpenAPI essentially solves this problem - we currently have that integration in place as swagger-jsdoc but I'm thinking a yaml approach will be nicer since it'll play better with openapi validator or whatever that thing is called

anyhow - moral of the story is that I want us to push for that design-first methodology; this has to be done before we do any more work on the server as it's a huge blocker in terms of decoupling frontend work from backend work

thaigillespie:
Got it, I'll make the changes for that on the front end. Sorry about that! And yeah I guess we should have gone over how the schema for the sign up form should look like first.

godwinpang:
no it's ok don't apologize; this is an iterative process for us as a team

no need to make the changes on the frontend - everything works rn so let's just roll with it 
we can create a ticket to make this stuff camelcase :))

godwinpang:
if it wasn't clear - please go ahead and merge this in

src/controllers/auth.controller.ts
AneeshJalan: Done in [this commit](https://github.com/HKN-UCSD/hkn-server/commit/4f7df3034bbcc36bcd5c58920eea8cd51eaf4fb2)

AneeshJalan:
Merging after pulling out the abstraction to send verification email.

src/index.ts
```diff
@@ -1,14 +1,28 @@
 import express from 'express';
 import admin from 'firebase-admin';
+import { firebase as client } from '@firebase/app';
+import '@firebase/auth';
```
godwinpang, line 4: hey aneesh, do you remember why this is being imported at the index.ts level?

godwinpang: @AneeshJalan

AneeshJalan: I felt that this may be used in some other services/controllers as well. No specific reason though for importing at the index.ts level.

PR #21, opened 2020-07-07
Added base AuthMiddleware and OfficerAuth and MemberAuth.

thaigillespie:
AuthMiddleware works very similarly to the frontend RBAC HOCs.
It's just that for AuthMiddleware, we use Firebase Admin SDK
instead of React Context.

OfficerAuth uses AuthMiddleware to be an Auth middleware that
checks if a user is an officer.

MemberAuth uses AuthMiddleware to be an Auth middleware that
checks if a user is a member.

### DIFF
--- a/src/middlewares/authentication/base/authentication.middleware.ts
+++ b/src/middlewares/authentication/base/authentication.middleware.ts
@@ -1,0 +1,34 @@
+import admin from 'firebase-admin';
+import { Request, Response, NextFunction } from 'express';
+
+import * as ERROR_MSG from '../../../constants/authErrorMessages';
+
+const AuthMiddleware = (permittedRoles: Array<String>) => (
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  const { token } = req.body;
+
+  admin
+    .auth()
+    .verifyIdToken(token)
+    .then(claims => {
+      const claimsKeyArr = Object.keys(claims);
+
+      if (claims != null) {
+        for (let i = 0; i < claimsKeyArr.length; i += 1) {
+          if (permittedRoles.includes(claimsKeyArr[i])) {
+            return next();
+          }
+        }
+
+        return next(new Error(ERROR_MSG.NO_PERMITTED_ROLES));
+      }
+    })
+    .catch(err => {
+      return next(err);
+    });
+};
+
+export default AuthMiddleware;
--- a/src/middlewares/authentication/base/authentication.middleware.ts
+++ b/src/middlewares/authentication/base/authentication.middleware.ts
@@ -1,6 +1,8 @@
 import { Request, Response, NextFunction } from 'express';
 import { verifyRole } from '../../../services/auth.service';
 
+import * as ERROR_MSG from '../../../constants/authErrorMessages';
+
 const AuthMiddleware = (permittedRoles: Array<string>) => async (
   req: Request,
   res: Response,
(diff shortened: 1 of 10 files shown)


src/middlewares/authentication/base/authentication.middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as ERROR_MSG from '../../../constants/authErrorMessages';
+
+const AuthMiddleware = (permittedRoles: Array<String>) => (
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  const { token } = req.body;
+
+  admin
```
godwinpang, line 13: let's rip this out into like auth.service.ts which exports a method called verifyRole or something - it'll make testing a lot easier.

thaigillespie: You mean the whole base AuthMiddleware? And what will verifyRole take as parameters? Sorry I'm just a little confused

src/middlewares/authentication/base/authentication.middleware.ts
```diff
@@ -0,0 +1,34 @@
+import admin from 'firebase-admin';
+import { Request, Response, NextFunction } from 'express';
+
+import * as ERROR_MSG from '../../../constants/authErrorMessages';
+
+const AuthMiddleware = (permittedRoles: Array<String>) => (
```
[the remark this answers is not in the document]
thaigillespie: You mean the whole AuthMiddleware function or verifyIdToken?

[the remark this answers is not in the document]
godwinpang: you can make authmiddleware async and use await inside

src/middlewares/authentication/base/authentication.middleware.ts
godwinpang: just wrapping the admin.VerifyIdToken stuff into a method that we own to decouple our controller from firebase to make testing easier

src/middlewares/authentication/base/authentication.middleware.ts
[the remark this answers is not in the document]
thaigillespie: If I use await like:

`await admin.auth().verifyIdToken(token) `

will I need to use .then and .catch after .verifyIdToken or not? I don't think so right? Since await should cover that already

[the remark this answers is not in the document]
godwinpang: `try {
    const claims = await admin.stuff
    verify stuff
} catch e {
    handle error
}`

See aneesh's pr for an example

src/middlewares/authentication/base/authentication.middleware.ts
thaigillespie: Oh so move the entire admin.verifyIdToken block to another file gotcha

godwinpang: you can only use the await keyword in an async function

src/middlewares/authentication/base/authentication.middleware.ts
```diff
@@ -1,34 +1,14 @@
...
-          }
-        }
-
-        return next(new Error(ERROR_MSG.NO_PERMITTED_ROLES));
-      }
-    })
-    .catch(err => {
-      return next(err);
-    });
+  await verifyRole(token, permittedRoles, next);
```
godwinpang, line 11: don't pass in next

godwinpang: instead you can have verifyRole return a boolean indicating success and act on that

thaigillespie: Oof sorry, I'll make it so that it returns true if user is authenticated and returns false if user is unauthenticated.

thaigillespie:
https://stackoverflow.com/questions/53940043/unhandledpromiserejectionwarning-this-error-originated-either-by-throwing-insid

I stumbled upon this link, which helped me decide what to do for the error throwing and error catching of verifyRole() and AuthMiddleware, respectively.

godwinpang:
if it wasn't clear - please go ahead and merge this in

godwinpang:
i like the documentation in the commit msg - would be good to have a concise version in the file as well :)

PR #23, opened 2020-07-10
Circleci project setup for hkn-server

thaigillespie:
So I did some basic set ups. Read through some parts of the CircleCI config reference (https://circleci.com/docs/2.0/configuration-reference/#steps) and looked at the ./circleci/config.yml file in hkn-web-portal to understand what is being done there. I just ended up copying hkn-web-portal's config.yml file because it has all that I need (I think).


src/index.ts
```diff
@@ -13,11 +13,6 @@ admin.initializeApp({
   databaseURL: config.dbURL,
 });
 
-
-console.log(config.clientApiKey)
```
godwinpang, line 17: sneaky i like it

godwinpang:
🔥 lgtm merge

src/index.ts
thaigillespie: LMAO they're not undefined I swear

PR #25, opened 2020-07-10
Added run_build and run_linter_prettier jobs for CircleCI.

thaigillespie:
run_build executes 'npm run build', while run_linter_prettier
executes 'npm run lint' then 'npx prettier --check ./src/**/*.ts.

run_build is dependent upon run_lint_prettier, in the sense that
run_build can only start if run_lint_prettier succeeds.


godwinpang:
uh oh looks like prettier isn't being run on some files

godwinpang:
dont think it has anything to do with your changes; just some things from the previous commits

thaigillespie:
yeah lmao I done goofed up

godwinpang:
no I think you're fine - looks like maybe @AneeshJalan doesn't have precommit hooks installed

thaigillespie:
Maybe, but I just ran 'npx prettier --write src/**/*.ts' anyway so should be good

thaigillespie:
This should be done now since I actually ran npx prettier --write :)))

godwinpang:
🔥

thaigillespie:
Should I merge it then or is there anything else I need to cover in this PR?

godwinpang:
mnerge

PR #29, opened 2020-07-20
Added joi middleware to provide a good way of validating req.body.

thaigillespie:
Since we have decided to use joi as the main library for validating
req.body (request body) in our back end code, I wrote this middleware
to make that task a lot less painful and to modularize the code better.

To use the middleware, import from the src/middlewares/validation/base
folder. Then, use it as a middleware on HTTP requests that require
the use of req.body.

Example:

app.post('/', ValidationMiddleware(schema), (req, res) => {
    res.json({
       Success: 'Success!'
    });
});


src/constants/ErrResponses.ts
```diff
@@ -6,3 +6,12 @@ export const USER_ALREADY_EXISTS =
...
   'There was an internal error. Please try again later.';
+
+// Auth Error Responses
+export const NO_PERMITTED_ROLES =
+  'The user does not have the required roles for this action.';
+export const USER_NOT_AUTHENTICATED = 'The user is unauthenticated.';
+
+// Validation Error Responses
+export const INVALID_REQUEST_BODY =
+  'The provided request body does not match with the schema it is supposed to follow.';
```
godwinpang, line 17: we should parameterize this so that we tell the caller what part doesn't match

src/constants/ErrResponses.ts
```diff
@@ -6,3 +6,12 @@ export const USER_ALREADY_EXISTS =
   'The provided email is aready linked to an account. Please check the email you have entered.';
 export const GENERIC_INTERNAL_ERROR =
   'There was an internal error. Please try again later.';
+
+// Auth Error Responses
+export const NO_PERMITTED_ROLES =
```
godwinpang, line 11: USER_NOT_AUTHORIZED

src/constants/ErrResponses.ts
thaigillespie: You mean like turn it into a function?

PR #30, opened 2020-07-23
Email integration using SendGrid

thaigillespie:


src/services/email.service.ts
```diff
@@ -0,0 +1,42 @@
+import * as SendGridMail from '@sendgrid/mail';
+import * as dotenv from 'dotenv';
+
+dotenv.config();
+SendGridMail.setApiKey(process.env.SENDGRID_API_KEY);
+
+/* Send an email to one or more recipients */
+const _sendEmails = async (
```
godwinpang, line 8: single Email?

src/services/email.service.ts
```diff
@@ -0,0 +1,42 @@
...
+      templateId,
+      cc,
+      bcc,
+      sendAt,
+      ...additionalFields
+    },
+    ...additionalEmailInfoObjs
+  ] = emailInfoArr;
+
+  try {
```
godwinpang, line 24: skip try catch if you're just gonna throw; add todo to add logging when we have that setup

.gitignore
```diff
@@ -4,3 +4,4 @@ dist
 .DS_Store
 .eslintignore
 .eslintcache
+./src/test.ts
```
godwinpang, line 7: just don't git add the file

src/services/email.service.ts
thaigillespie: Old comment before I changed it from sendAnEmail to _sendEmails lol sorry

godwinpang: sorry - I meant it would be more useful to expose the single email func

src/services/email.service.ts
```diff
@@ -0,0 +1,42 @@
...
+
+dotenv.config();
+SendGridMail.setApiKey(process.env.SENDGRID_API_KEY);
+
+/* Send an email to one or more recipients */
+const _sendEmails = async (
+  emailInfoArr: SendGridMail.MailDataRequired[]
+): Promise<void> => {
+  const [
+    {
```
godwinpang, line 17: I don't see why we have to destructure here, aren't you extracting and reassembling the fields into the same obj? what might be useful is instead of exposing the SendGridMail as the param to the public method, we create our own mail struct/interface/type or whatever it's called in ts and translate that to sendgridmail.mailxxx

that way we don't break the email abstraction

thaigillespie: The destructuring was like a way for me to "expose" the fields I guess... And yes I can write a type interface to use instead of having to use SendGridMail.MailDataRequired[] for emailInfoArr.

thaigillespie: Where should I put this type definition file though? You want me to make a whole new folder for it?

godwinpang: I'm also foreseeing the main usecase being sending a single email bcc-ed to a lot of people

godwinpang: for now each service can be it's own folder inside services with service + types in two files

godwinpang: idr the difference between interfaces and types in ts so you'll have to read up

thaigillespie: I'll restrict the interface's properties to just what we listed in the docs + 'from", is that okay? Otherwise if we just take all of what SendGridMail.MailDataRequired has then we might as well just use their data type.

src/services/email/email.definition.ts
```diff
@@ -0,0 +1,9 @@
+export interface EmailInfo {
+  to: Array<string>;
+  from: string;
```
godwinpang, line 3: emailService should fill this in with whatever email we're sending from

src/services/email/email.definition.ts
```diff
@@ -0,0 +1,9 @@
+export interface EmailInfo {
+  to: Array<string>;
```
godwinpang, line 2: only one of cc, bcc, to is required, not necessary that to is always filled in (unless sendgrid requires it to be)

src/services/email/email.definition.ts
godwinpang: aka as a developer it would be nice not to have the hkn email in multiple places around the codebase

src/services/email.service.ts
godwinpang: could perhaps also use a typealias https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases

src/services/email/email.definition.ts
thaigillespie: Yeah that's true. To be frank cc, bcc and to are all optional properties of SendGridMail's type

thaigillespie:
I can just fix it up real quick haha. Pretty close to being done tbh

src/services/email/email.definition.ts
thaigillespie: Well actually, 'from' here only defines what you need to pass into the emailInfo object. As in the object needs to have 'from' (SendGrid requires 'from' as well, so we'll need to have it). What we can do is just to have an env var or something for the hkn email, then pass it into 'from' when we call the public email functions in the future.

src/services/email/email.definition.ts
thaigillespie: I'll make 'to' optional as well

src/services/email.service.ts
thaigillespie: Yep. In our case, it's the same whether we use interface or type alias. There are differences between them but for the things we are doing here we don't need to worry about it just yet.

src/services/email/email.definition.ts
godwinpang: in general it's good practice to load env vars only once - should load it in email.service and hide it from the user

thaigillespie: Yeah that's fine as well. I'll do that. Though we will have to keep 'from' property's type definition.

godwinpang: why?

thaigillespie: Because SendGrid's send() function requires it. Our custom interface has to have it or else it will give an error. I have tried taking it out already, believe me. And if I do take 'from' out, and try to insert 'from' into the EmailInfo objects, then it will also give an error, so there is no way around it.

godwinpang: ah gotcha - not sure why reinserting from doesn't work but I'll take your word on that

thaigillespie: I can demo tomorrow if you would like. Anything else you want me to fix? I think this should be ready for merging now.

thaigillespie: Also, I think reinserting 'from' doesn't work because once I have declared the specific custom EmailInfo type for the params, then it uses that to match types when u pass the EmailInfo object as a param to SendGrid's send(). Meaning that even if you reinsert 'from', its type still doesn't have a 'from' property so it gives an error.

src/services/email/email.service.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/*
+ * Send one or more emails to one or more respective recipients
+ * for each of those emails
+ *
+ * TODO: Add logging when there is a logging set up
+ */
+const _sendEmails = async (emailInfoArr: Array<EmailInfo>): Promise<void> => {
+  const emailsToSend = emailInfoArr.map(emailInfo => {
+    emailInfo.from = senderEmail;
```
godwinpang, line 18: hm from is still required so do callers leave it empty?

thaigillespie: Oh shoot you're right. Uhhhh if callers leave it empty it's gonna give an error in the text editor, I just tested it.

thaigillespie: I think I might have a solution to this tho.

godwinpang: It’s ok - I’d rather we setup error handling and logging soon, both of which would be rather straightforward 
We can discuss this tmrw

thaigillespie: I'm also working on the error handling middleware. I'll make a PR as soon as I think I'm done with it.

PR #32, opened 2020-07-25
Added error handler middleware to the end of src/index.ts.

thaigillespie:


src/middlewares/errorHandler/errorHandler.middleware.ts
```diff
@@ -0,0 +1,12 @@
+import { Request, Response, NextFunction, ErrorRequestHandler } from 'express';
+
+const ErrorHandler: ErrorRequestHandler = (
+  err: Error,
```
AneeshJalan, line 4: Is Error a standard typescript type or should that also be imported from 'express'?

thaigillespie: I think it's a Typescript standard type.

src/middlewares/errorHandler/errorHandler.middleware.ts
```diff
@@ -0,0 +1,12 @@
...
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  res.status(500).send({ err, testing: 'testing!' });
```
godwinpang, line 9: take out the testing lol

thaigillespie: lol I didn't even notice, thanks for pointing it out.

PR #33, opened 2020-07-26
Added rate limiting to all endpoints so we don't get DDOS-ed.

thaigillespie:


src/index.ts
```diff
@@ -1,12 +1,21 @@
...
+
 import { config } from './config';
 import { UserRouter } from './routers/user.router';
 import { DocsRouter } from './routers/docs.router';
 import { AuthRouter } from './routers/auth.router';
+import rateLimit from 'express-rate-limit';
+
+const limiter = rateLimit({
+  windowMs: parseInt(process.env.RATE_LIMIT_TIMEFRAME, 10),
+  max: parseInt(process.env.RATE_LIMIT_MAX_CONNS, 10),
```
godwinpang, line 17: maxConns is a misnomer: should be maxRequests

PR #35, opened 2020-07-28
Postgres POC 

godwinpang:
Proof of concept of typeorm onboarding complete with typedi for DI testing (will make life easier in the future in terms of swapping services out once I setup interfaces.)

[routing-controllers](https://github.com/typestack/routing-controllers) is great - if you check event.controller.ts you will see that we no longer have to deal with param parsing + casting, 404 error responses on null etc.

There's a sample jest test suite in the tests folder along with their mocks so you can see what testing with typedi looks like with mocked modules. technically this is the same as mocked functions in jest rn, but having a di framework scales a little better.

We could roll out DI ourselves without a framework and manually inject through constructor, but routing-controllers doesn't play too nicely with manual DI so I thought we might as well onboard typedi


godwinpang:
when reviewing, please mainly focus on event.service.ts, event.controller.ts, entities/, and everything in tests

src/index.ts
```diff
@@ -29,18 +39,30 @@ client.initializeApp({
...
-app.use(limiter);
+    const port = process.env.PORT || 3001;
 
-app.use('/api/user', UserRouter);
-app.use('/docs', DocsRouter);
-app.use('/api/auth', AuthRouter);
+    app.use(express.json());
+    app.use(express.urlencoded({ extended: true }));
+    app.use(limiter);
+    app.use(ErrorHandler);
```
thaigillespie, line 59: Error handler middleware should be put at the end of all the app.use() calls, if you do plan to keep it there. I think you want to remove it though?

src/index.ts
```diff
@@ -1,7 +1,9 @@
+import 'reflect-metadata'; // shim required for routing-controllers
+import 'module-alias/register';
 import express from 'express';
 import cors from 'cors';
```
thaigillespie, line 4: It doesn't look like we're using cors anymore in this file so we can probably remove this import statement.

src/tests/event.controller.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+describe('EventController', () => {
+  beforeEach(() => {
+    const mockEventService = new MockEventService(new Map());
+    const mockAppUserService = new MockAppUserService();
+    Container.set(EventService, mockEventService);
+    Container.set(AppUserService, mockAppUserService);
+  });
+
+  describe('createEvent', () => {
```
thaigillespie, line 18: What does this test currently do exactly? As far as I can tell it's only checking if true is equal to true.

src/entities/Event.ts
```diff
@@ -0,0 +1,75 @@
...
+} from 'typeorm';
+import { AppUser } from './AppUser';
+import { Attendance } from './Attendance';
+import { RSVP } from './RSVP';
+
+export enum EventType {
+  PROFESSIONAL = 'professional',
+  SOCIAL = 'social',
+  TECHNICAL = 'technical',
+  MENTOSHIP = 'mentorship',
```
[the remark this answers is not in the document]
godwinpang: good catch - thanks

src/entities/Attendance.ts
```diff
@@ -0,0 +1,33 @@
...
+import { AppUser } from './AppUser';
+import { Event } from './Event';
+
+// TODO composite index on attendee + event
+@Entity()
+export class Attendance extends BaseEntity {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @ManyToOne(() => AppUser)
```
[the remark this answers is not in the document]
godwinpang: an attendance is defined as this user attended this event. since a user can attend multiple events, it's manytoone

src/index.ts
godwinpang: ah hm - that makes sense. not sure what I want to do yet

src/index.ts
godwinpang: yep - will do. have a lot of stuff I haven't had the time to cleanup yet

src/tests/event.controller.test.ts
godwinpang: yeah - it only check true is equal to true. it was mainly a POC to demonstrate how to mock stuff with typedi

godwinpang:
good point about the writeup - will get started on it

godwinpang:
@thaigillespie @AneeshJalan  is this good - can i merge? sorry it's such a big pr

src/controllers/index.ts
```diff
@@ -0,0 +1,4 @@
+import { EventController } from './EventController';
+
+export { EventController };
```
thaigillespie, line 3: Why two export statements here? Don't you need only one?

src/entities/Event.ts
```diff
@@ -0,0 +1,78 @@
...
+  })
+  status: string;
+
+  @ManyToMany(() => AppUser, { eager: true })
+  @JoinTable()
+  hosts: AppUser[];
+
+  @OneToMany(
+    () => Attendance,
+    attendance => attendance.event
```
thaigillespie, line 57: Just curious on what this line means

src/controllers/index.ts
godwinpang: there's a reference to EventController within index.ts as im exporting the array of controllers to be used in index.ts and the docs stuff

src/entities/Event.ts
godwinpang: https://github.com/typeorm/typeorm/blob/master/docs/many-to-one-one-to-many-relations.md

src/entities/InductionClass.ts
```diff
@@ -0,0 +1,16 @@
...
+@Entity()
+export class InductionClass {
+  @PrimaryColumn()
+  quarter: string;
+
+  @Column()
+  name: string;
+
+  @Column('date')
+  startDate: string;
```
thaigillespie, line 12: What does the start date and end date indicate here? I thought these induction classes are permanent?

godwinpang: ah good call - i meant this to be induction period, so FA19 would be something like september 1st 2019 - whenever

thaigillespie: Oh okay, that's fine with me then.

src/payloads/Event.ts
```diff
@@ -0,0 +1,69 @@
...
+  readonly fbURL: string;
+
+  @IsUrl()
+  @IsOptional()
+  readonly canvaURL: string;
+}
+
+export class EventRequest extends BaseEventPayload {
+  @IsInt({ each: true })
+  readonly hosts: number[];
```
thaigillespie, line 54: I thought ids for AppUsers are strings? Or are they just numbers in string form?

src/services/AppUserService.ts
```diff
@@ -0,0 +1,18 @@
+import { AppUser } from '@Entities';
+import { injectable } from 'tsyringe';
+import { Any, getRepository } from 'typeorm';
+import { AppUserServiceInterface } from './interfaces/AppUserServiceInterface';
+import { Repository } from 'typeorm';
```
thaigillespie, line 5: Why two separate import statements from 'typeorm'?

src/services/AppUserService.ts
```diff
@@ -0,0 +1,18 @@
...
+@injectable()
+export class AppUserService implements AppUserServiceInterface {
+  private appUserRepository: Repository<AppUser>;
+
+  constructor() {
+    this.appUserRepository = getRepository(AppUser);
+  }
+
+  getMultipleAppUsers(ids: number[]): Promise<AppUser[]> {
+    return this.appUserRepository.find({ id: Any(ids) });
```
thaigillespie, line 16: Just curious on how this line works

src/services/AppUserService.ts
godwinpang: because i relied on autoimport :)))

src/services/AppUserService.ts
godwinpang: https://orkhan.gitbook.io/typeorm/docs/find-options - maps to ANY in sql
python equivalent is sth like appUserTable.filter(n -> n in ids)

src/payloads/Event.ts
godwinpang: postgres uses autoincrement integers instead of uuids, that's why all ids are numbers

godwinpang: aaaaaaaaaaaaaand you're right I trolled - good catch @thaigillespie

thaigillespie: haha npnp

PR #36, opened 2020-08-02
Added morgan for request logging and winston for internal logging.

thaigillespie:


src/logging.ts
```diff
@@ -0,0 +1,11 @@
+import winston from 'winston';
+import dotenv from 'dotenv';
+
+dotenv.config();
```
godwinpang, line 4: dotenv.config should be called only once - the idea is that we have a config in config.ts and call it there. 
from there on out, nothing else will directly access process env. they will instead access the config

thaigillespie: Shoot I used dotenv.config() outside of config.js in a couple more places in the codebase. I'll remove them in this PR then?

godwinpang: it's ok - do it after postgres merges in

godwinpang:
future todo is to reference: https://github.com/UCSDTESC/Check-in/blob/master/src/server/config/Logging.ts

PR #42, opened 2020-08-10
Docs and cleanup

godwinpang:
Heads up this is a fat PR because I didn't commit as much as I
should have.

Services are now directly injected via @Inject(Service) instead of
registering them with an interface. While DI against an interface is the
"right" thing to do, we treat our services as a unit of business logic,
and thus there is no need to swap them out with DI interfaces. This
makes development a little easier as there's no need to go through the
ritual of writing interfaces in order to develop a single use business
logic service.

For the use cases of swapping out email/S3/auth services, we can do that on a
case by case basis and onboard the interface complexity as needed.

Services/Controllers are now @singleton instead of @Injectable. Since
they are all stateless bags of functions, we can share a single instance
globally.

Repositories are now injected properly so we can test our services a
little better. without having to resolve to using jest.mock()/jest.fn().
Unfortunately due to the nature of typeorm's getRepository call, we are
forced to register our repositories using a token and resolving it via
said tokens instead of being able to do something like @injectRepository.
All repositories and their factories are in repositories/index.ts.
As a result, repositories now have to be loaded in
loaders/RepositoryLoader.ts.

Services have been stripped of their dependencies on payloads. They now
directly operate on entities. Entity to payload conversion and vice
versa has been offloaded to a new type of class called mappers. All
these mappers do is translate payloads into entities, and entities into
payloads.


src/entities/InductionClass.ts
```diff
@@ -1,16 +1,25 @@
...
 
+  // e.g. "Zeta Kappa"
   @Column()
   name: string;
 
+  // Starting date of the induction cycle for said class
   @Column('date')
   startDate: string;
 
+  // Starting date of the induction cycle for said class
```
thaigillespie, line 22: Fix comment typo

src/payloads/AppUser.ts
```diff
@@ -0,0 +1,6 @@
+import { IsInt } from 'class-validator';
+
+export class AppUserPKPayload {
```
thaigillespie, line 3: What's pk here?

thaigillespie:
Let me change the email thing tmr tho.

src/payloads/AppUser.ts
godwinpang: primary key

thaigillespie:
Email column changed from @Column to @Column({ primary: true })

thaigillespie:
This should be good to go now. I'm pretty sure when we have multiple primary columns in orm, it just makes a composite primary key from that... Or at least that's what the docs are saying https://orkhan.gitbook.io/typeorm/docs/entities#entity-columns

godwinpang:
hey @thaigillespie thanks for fixing this. what do you think about email being unique key instead of making a composite primary key? The reason I bring this up is that with a composite pk we technically could have something that looks like (id1, email1); (id1, email2). This isn't possible with autogenerated ids, but a composite PK also brings along the problem of requiring both id and email being embedded as foreign keys in other tables when the id is sufficient to identify the user

maybe we could make email not null, unique, and put an index on it? that way we keep the id as a pk and simplify the foreign keys on other tables.  let me know what you think

thaigillespie:
I think that make email unique instead of making a primary column is better because like you said, id is already more than enough to identify a user within the db.

thaigillespie:
wait so let's change that then? I think based on what u said (no null, unique, put index), I just need @Column({ unique: true }) right?

PR #43, opened 2020-08-10
Add compression middleware.

godwinpang:


godwinpang:
ngl idk - just went off this

https://expressjs.com/en/advanced/best-practice-performance.html#use-gzip-compression

thaigillespie:
Well if express says it's good then it's most likely good then.

PR #44, opened 2020-08-13
Event sign in endpoint (Ready for review)

thaigillespie:


thaigillespie:
"I still don't know what to do if the AppUser's role is an affiliate role
(so inductee, member, officer, anything that is not guest). Might just
treat those roles the same way as how I handle attendance register for
non-affiliates." (Latest commit's message)

If the user signing in to an event is an affiliate, then what do we do? So far, it's
"Return HTTP error", but I think it's better to just do what I did for when the user 
is a non-affiliate signing up for the first time and non-affiliate signing up for second or more time. 

Also, can I have some pointers on how to do testing? Maybe some kind of workflow or smth idk.

godwinpang:
> If the user signing in to an event is an affiliate, then what do we do? So far, it's
> "Return HTTP error", but I think it's better to just do what I did for when the user
> is a non-affiliate signing up for the first time and non-affiliate signing up for second or more time.
> 

I don't follow - can you elaborate

godwinpang:
> Also, can I have some pointers on how to do testing? Maybe some kind of workflow or smth idk.

if you have heroku cli and access to the heroku db you can run this: `heroku pg:psql -a hkn-server-dev`

godwinpang:
for events I just ran it against my postman collection since I had CRUD on events so I could do stuff like Create event then Get all events

thaigillespie:
> 
> 
> > If the user signing in to an event is an affiliate, then what do we do? So far, it's
> > "Return HTTP error", but I think it's better to just do what I did for when the user
> > is a non-affiliate signing up for the first time and non-affiliate signing up for second or more time.
> 
> I don't follow - can you elaborate

So idk if you remember in the pseudocode you wrote in a thread a few days ago, it says if the user is an affiliate, then return some kind of HTTP error. But I'm thinking just handle that case like other non-affiliate cases, so get the AppUser row for that user, then make an Attendance entity for them for a specified event.

godwinpang:
an affiliate wouldn't have to post with a body though right? they can just post to the endpoint with an empty body and the authentication token since they're already logged in. We'll be able to grab the user id from the token

godwinpang:
so to elaborate, what I meant was to send an HTTP error if an affiliate is trying to signin to an event through the non-affiliate form. if they're not signing in through the non-affiliate form and are instead signing in through their internal page where they're logged in, then they will post to the endpoint with their authentication token

thaigillespie:
But isn't the token in req.body?

godwinpang:
that was the original plan until i learned that about https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

you're supposed to jut your token into headers

thaigillespie:
So something like this to have access to the token then? https://github.com/typestack/routing-controllers#inject-request-header-parameters

godwinpang:
https://github.com/typestack/routing-controllers#currentuser-decorator

leave this for another pr - out of scope

godwinpang:
or maybe another middleware idk

thaigillespie:
I was gonna check if the token exists. If it does, then the user is signing in from their internal page. If not, then they're signing in from the non-affiliate page. Maybe there's a better way of doing this?

godwinpang:
wait does this work if i reply through email

that's definitely a valid approach - just wondering if we could yank out
the token=>user into a middleware (either something routing-controller
already provides or we roll our own) since there's no use case in which you
just look at the token without verifying it. still need to evaluate
their @CurrentUser decorator to support a development authentication
strategy that doesn't depend on firebase tokens.

thaigillespie:
I see that makes sense. And yes send by mail works @godwinpang

thaigillespie:
I can look more into @CurrentUser decorator and incorporate it into the PR to get this done. If an affiliate has a token, then verify it; if not, then we send an HTTP error.

godwinpang:
no - let's leave it for a separate PR I anticipate that there's a good
amount of complexity

godwinpang:
just keep the scope for this PR as serving non-affiliate event signups

src/entities/Attendance.ts
```diff
@@ -10,21 +10,18 @@ import { Event } from './Event';
...
   @ManyToOne(() => AppUser)
   officer: AppUser;
 
-  @ManyToOne(() => Event)
+  @ManyToOne(() => Event, { primary: true })
   event: Event;
 
   // num_hours
-  @Column('float')
+  @Column('float', { nullable: true })
```
godwinpang, line 24: out of scope but maybe duration can be a computed value of end time - start time? ik typeorm can autogenerate a creation time column. food for thought

godwinpang: and tbf this is mb - i just jammed duration in there without much thought

src/mappers/AppUserMapper.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  async requestToExistingEntity(
+    appUserRequest: EventSignInRequest,
+    appUserId: number
+  ): Promise<AppUser | undefined> {
+    const appUserObj: AppUser = appUserRequest as AppUser;
+    appUserObj.id = appUserId;
+
+    const appUser: AppUser = await this.appUserRepository.preload(appUserObj);
+    if (appUser == undefined) {
```
godwinpang, line 45: can probably just return appUser

src/payloads/AppUser.ts
```diff
@@ -1,6 +1,43 @@
...
+  readonly lastName: string;
+
+  @IsEmail()
+  readonly email: string;
+
+  @IsString()
+  readonly major: string;
+}
+
+export class EventSignInRequest extends BaseAppUserPayload {}
```
godwinpang, line 23: hm thoughts on moving this to event payloads? feels a little weird to have eventreq/res in the appuser file. we can import the appuserpayload in the event file :)

src/services/EventService.ts
```diff
@@ -50,4 +56,20 @@ export class EventService {
...
+   * the passed-in AppUser entity, then stores that Attendance entity to the
+   * Attendance table.
+   *
+   * @param eventId
+   * @param appUser
+   */
+  async registerAttendance(eventId: number, appUser: AppUser): Promise<Attendance> {
+    const event = await this.eventRepository.findOne({ id: eventId });
+    const attendance = this.attendanceService.createAttendance(event, appUser);
+    const savedAttendance = await this.attendanceService.saveAttendance(attendance);
```
godwinpang, line 71: I don't have an answer to this question, but I think the current implementation allows for multiple sign-ins of the same user to the same event. they'll be stored in the database as the same thing (aka no problem of duplicated data) but there's no error code. perhaps in the future there is merit to "denying" a user multiple signins to the same event just from a UX perspective.

godwinpang: that being said - this current implementation works and is good enough - just marking for historical purposes

src/controllers/EventController.ts
```diff
@@ -75,4 +81,38 @@ export class EventController {
 
     return this.eventMapper.entityToResponse(deletedEvent);
   }
+
+  @Post('/:eventID/signin')
+  @ResponseSchema(Attendance)
```
godwinpang, line 86: probably can drop since this returns null

src/controllers/EventController.ts
```diff
@@ -75,4 +81,38 @@ export class EventController {
...
+  async signInToEvent(
+    @Param('eventID') eventID: number,
+    @Body() appUserRequest: EventSignInRequest
+  ): Promise<Attendance> {
```
godwinpang, line 90: promise void

src/services/AttendanceService.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@singleton()
+export class AttendanceService {
+  private attendanceRepository: Repository<Attendance>;
+
+  constructor(@inject(AttendanceRepositoryToken) attendanceRepository: Repository<Attendance>) {
+    this.attendanceRepository = attendanceRepository;
+  }
+
+  createAttendance(event: Event, attendee: AppUser): Attendance {
```
godwinpang, line 15: what do you think about renaming this to registerAttendance(event, attendee) and calling attendancerepo.save within the function? I don't see a use case where we createattendances and dont persist them to the db

src/payloads/AppUser.ts
thaigillespie: Yeah we can def do that

src/controllers/EventController.ts
thaigillespie: I can make an AttendanceMapper and stuff and then have a proper AttendanceResponsePayload or smth like that. And yeah I can also just do what you said here

src/mappers/AppUserMapper.ts
thaigillespie: You mean no null check?

godwinpang: oh that's == undefined not === undefined lol

preload - Creates a new entity from the given plain javascript object. If the entity already exists in the database, then it loads it (and everything related to it), replaces all values with the new ones from the given object, and returns the new entity. The new entity is actually an entity loaded from the database with all properties replaced from the new object.
Note that given entity-like object must have an entity id / primary key to find entity by. Returns undefined if entity with given id was not found.

godwinpang: can probably just return it since preload returns undefined

src/entities/Attendance.ts
thaigillespie: yeah that's what duration should be

src/services/EventService.ts
thaigillespie: I think right now I allow updates to the Attendance entity. Since if the same user signs in again, it'd be the same composite PK so they'd just be updating their already existing Attendance entity. I think maybe we can move the updating to PUT (?) since at initial sign in, Attendance won't have values for fields like officer and duration.

godwinpang: yeah i agree with the first part - probably PATCH makes more sense than PUT for updating in this specific case. what I was thinking was just that we might not want users to be able to make multiple POST calls to the attendance endpoint

src/services/AppUserService.ts
```diff
@@ -18,4 +32,34 @@ export class AppUserService {
...
+   * Get an existing AppUser by their email address.
+   *
+   * @param email The email used to look for the corresponding AppUser.
+   *
+   */
+  getAppUserByEmail(email: string): Promise<AppUser> {
+    return this.appUserRepository.findOne({ email });
+  }
+
+  async saveNonAffiliate(appUserRequest: EventSignInRequest): Promise<AppUser | undefined> {
```
godwinpang, line 46: thoughts on doing the mapping in the controller instead? that way services dont have to know about payloads :)

src/services/AttendanceService.ts
```diff
@@ -0,0 +1,34 @@
...
+    try {
+      await this.attendanceRepository.insert(newAttendance);
+    } catch {
+      return undefined;
+    }
+
+    return newAttendance;
+  }
+
+  async getAttendanceByEventUser(event: Event, attendee: AppUser): Promise<Attendance | undefined> {
```
godwinpang, line 29: unused function?

src/services/AppUserService.ts
godwinpang, line 46: hm I think it would be nice if services didn't need to know about payloads. what if we can just have a requestToAppUser in the mapper? then saveNonAffiliate can just take in AppUser instead of AppUserRequest

thaigillespie: Yeah I can do that. We already have requestToNewEntity which is basically requestToAppUser since it just creates a new AppUser entity (with no saves).

src/controllers/EventController.ts
```diff
@@ -75,4 +81,35 @@ export class EventController {
...
+    @Body() appUserRequest: EventSignInRequest
+  ): Promise<Attendance | undefined> {
+    const { email } = appUserRequest;
+    const appUserFromEmail = await this.appUserService.getAppUserByEmail(email);
+    const appUserToSave = await this.appUserMapper.requestToEntityByEmail(
+      appUserFromEmail,
+      appUserRequest
+    );
+
+    if (appUserToSave === undefined) {
```
godwinpang, line 99: not sure why this check is needed

src/controllers/EventController.ts
```diff
@@ -75,4 +81,35 @@ export class EventController {
...
+  @Post('/:eventID/signin')
+  @OnUndefined(409)
+  @ResponseSchema(Attendance)
+  async signInToEvent(
+    @Param('eventID') eventID: number,
+    @Body() appUserRequest: EventSignInRequest
+  ): Promise<Attendance | undefined> {
+    const { email } = appUserRequest;
+    const appUserFromEmail = await this.appUserService.getAppUserByEmail(email);
+    const appUserToSave = await this.appUserMapper.requestToEntityByEmail(
```
godwinpang, line 94: hm a little weird to have mapper take in an entity if the mapping layer translates requests to entities can't we just pass in the request to the mapper and have it do the getAppUserByEmail stuff?

thaigillespie: Is it okay for mappers to use services then? If that's the case I'll change

godwinpang: I'd say mappers can use services yeah - mappers and controllers sit at the same layer of req/res parsing stuff. if controllers can use services then mappers should be able to do so as well.

thaigillespie: Sounds good

godwinpang: you bring up an interesting point though - if mappers are at the same layer as controllers then should they directly rely on repositories? (this is directed more at eventmapper)

lmk what you think and maybe we'll go ahead and change eventmapper to rely on services instead

src/controllers/EventController.ts
thaigillespie: Removed

src/controllers/EventController.ts
thaigillespie: I think if we don't rely directly on repositories for mappers, then we'll have to add more CRUD operations to services. If we do, however, I feel like it might be a violation layer because services already use repositories directly, and if mappers are at the same level as controllers then mappers are at a higher layer than services.

godwinpang:
Hm yeah that’s fair - I think another pro of this is that the dependency becomes very explicit: controllers/mappers => services => repositories. That makes for a nice layered architecture.

On the other hand, it also means that a service would have to take in a raw request instead of just an entity, which breaks consistency of having services not having to know about payloads.

thaigillespie:
Not sure myself really... Though at least the code works as intended. Other than design discussions, functionality wise this is good to go.

godwinpang:
true - sounds good. I'll create an issue and we can discuss the role of mappers there. Feel free to merge this in and we can come back to clean up later

PR #46, opened 2020-08-17
Attendance payload mapper

thaigillespie:


src/controllers/EventController.ts
```diff
@@ -93,7 +93,7 @@ export class EventController {
   @ResponseSchema(AttendanceResponse)
   async signInToEvent(
     @Param('eventID') eventID: number,
-    @Body() appUserRequest: EventSignInRequest
+    @Body() appUserRequest: AppUserEventRequest
   ): Promise<Attendance | undefined> {
```
godwinpang, line 97: Promise <AttendanceResponse | undefined>

src/payloads/AppUser.ts
```diff
@@ -1,20 +1,35 @@
...
   readonly firstName: string;
 
   @IsString()
   readonly lastName: string;
 
+  @IsString()
+  readonly major: string;
+}
+
+// If this class is empty, then it won't be decorated by class-validator, which will break docs
```
godwinpang, line 20: for sth like this you could also have Response directly extend Request - no need to change just a suggestion

src/payloads/Attendance.ts
```diff
@@ -1,24 +1,22 @@
...
-  @ValidateNested({ each: true })
-  @IsInstance(AppUser)
-  readonly attendee: AppUser;
+  @IsInstance(AppUserEventResponse)
+  readonly attendee: AppUserEventResponse;
 
-  @ValidateNested({ each: true })
-  @IsInstance(Event)
-  readonly event: Event;
+  @IsInstance(EventAttendanceResponse)
```
godwinpang, line 9: i don't think this is good because the endpoint is /events/:eventID/signin, so event shouldn't be in the payload for the request. it's fine if it's in the response but it it's not in the request then it shouldn't be in the BasePayload

src/payloads/AppUser.ts
godwinpang: honestly though - cleanest way might just be to screw all inheritance and duplicate fields between req/res

src/payloads/Attendance.ts
```diff
@@ -1,24 +1,22 @@
...
-import { AppUser, Event } from '@Entities';
+import { IsOptional, IsBoolean, IsNumber, IsInstance } from 'class-validator';
+import { AppUserEventResponse } from './AppUser';
+import { EventAttendanceResponse } from './Event';
 
 abstract class BaseAttendancePayload {
-  @ValidateNested({ each: true })
-  @IsInstance(AppUser)
-  readonly attendee: AppUser;
+  @IsInstance(AppUserEventResponse)
```
godwinpang, line 6: also feels a little weird to have response in the base payload esp since you'll possibly have attendance request later which would also extend BaseAttendancePayload

src/controllers/EventController.ts
godwinpang: I'm happy to merge after you fix this

godwinpang:
just to clarify - we can fix the inheritance stuff in a later PR feel free to merge in after fixing Promise

src/controllers/EventController.ts
thaigillespie: Won't I need a mapper if I do this?

godwinpang: yeah - if you look at https://github.com/HKN-UCSD/hkn-server/blob/eventSignInEndpoint/src/controllers/EventController.ts all the entities are mapped to responses explicitly

src/payloads/AppUser.ts
thaigillespie: I agree, it will be longer, but probably more clarity and we won't have to deal with the problems I had

src/payloads/Attendance.ts
thaigillespie: Yep agreed. I removed BaseAttendancePayload completely and just have AttendanceResponse by itself now

PR #49, opened 2020-08-19
Rbac and auth

thaigillespie:
Not quite done yet


src/services/AuthenticationService.ts
```diff
@@ -0,0 +1,52 @@
...
+  constructor(@inject(AppUserService) appUserService: AppUserService) {
+    this.appUserService = appUserService;
+  }
+
+  // Not sure if Promise<Object> or Promise<admin.auth.DecodedIdToken> is better...
+  async firebaseVerifyIdToken(token: string): Promise<AppUser | undefined> {
+    try {
+      const tokenResult = await admin.auth().verifyIdToken(token);
+
+      if (tokenResult != null) {
```
godwinpang, line 20: u can do if (!tokenResult) {return undef} to reduce 1 layer of nesting

src/services/AuthenticationService.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  async localVerifyIdToken(token: string): Promise<AppUser | undefined> {
+    const id = parseInt(token, 10);
+    return await this.appUserService.getAppUserById(id);
+  }
+
+  async verifyToken(mode: string, token: string): Promise<AppUser | undefined> {
+    if (mode === 'development') {
```
godwinpang, line 40: i think we can just hard code a check on the Config here

src/services/AuthenticationService.ts
godwinpang: also good to make these private

src/services/AuthenticationService.ts
godwinpang: instead of having to pass it in

src/config.ts
```diff
@@ -25,6 +26,7 @@ const {
   FIREBASE_DATABASE_URL,
   FIREBASE_CLIENT_ID,
   FIREBASE_CLIENT_API_KEY,
+  CUSTOM_ENV_MODE,
```
godwinpang, line 29: probably like DEV_AUTH or sth

src/middlewares/authentication/AuthenticationMiddleware.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  async use(request: any, response: any, next?: (err?: any) => any): Promise<any> {
+    const { headers } = request;
+    const token = headers['authorization'];
+    const { customEnvMode } = config;
+
+    try {
+      const tokenResult = await this.authenticationService.verifyToken(customEnvMode, token);
+
+      if (tokenResult !== undefined) {
```
godwinpang, line 24: can probably drop the else and do if (tokenResult === undef) {return next(blah)} return next()

src/middlewares/authorization/AuthorizationMiddleware.ts
```diff
@@ -0,0 +1,31 @@
...
+    this.permittedRoles = permittedRoles;
+  }
+
+  async use(request: any, response: any, next?: (err?: any) => any): Promise<any> {
+    const { permittedRoles } = this;
+
+    try {
+      // How to get user id from token from here? If we use verifyToken from AuthenticationService then it's
+      // basically doing both authentication and authorization for actions users can do after logging in.
+      //const isAuthorized = await this.authorizationService.hasSufficientRole(this.permittedRoles, );
```
godwinpang, line 26: what's wrong with verifyToken + hasSufficientRole? I think that looks good

thaigillespie: Oh alright, I'll go with that then

src/constants/ErrResponses.ts
```diff
@@ -9,6 +9,7 @@ export const GENERIC_INTERNAL_ERROR = 'There was an internal error. Please try a
 // Auth Error Responses
 export const USER_NOT_AUTHORIZED = 'The user is not authorized for this action.';
 export const USER_NOT_AUTHENTICATED = 'The user is unauthenticated.';
+export const INVALID_ROLE = 'The user has an invalid role.';
```
godwinpang, line 12: https://github.com/typestack/routing-controllers#throw-http-errors

src/middlewares/authorization/AuthorizationFactory.ts
```diff
@@ -0,0 +1,65 @@
...
+  constructor(
+    @inject(AuthenticationService) authenticationService: AuthenticationService,
+    @inject(AuthorizationService) authorizationService: AuthorizationService
+  ) {
+    this.authenticationService = authenticationService;
+    this.authorizationService = authorizationService;
+  }
+
+  getAuthorizationMiddleware(permissionLevel: string): AuthorizationMiddleware {
+    const { authenticationService, authorizationService } = this;
```
godwinpang, line 22: jw, does eslint complain if you don't do destructuring?

src/controllers/EventController.ts
```diff
@@ -1,8 +1,17 @@
...
+  Delete,
+  Body,
+  OnUndefined,
+  UseBefore,
+} from 'routing-controllers';
 import { singleton, inject } from 'tsyringe';
 import { ResponseSchema } from 'routing-controllers-openapi';
 
-import { Event } from '@Entities';
+import { Event, AppUserRole } from '@Entities';
```
godwinpang, line 14: is this an enum?

godwinpang: didnt see it defined anywhere

src/services/AuthorizationService.ts
```diff
@@ -0,0 +1,17 @@
...
+import { AppUser } from '@Entities';
+
+@singleton()
+export class AuthorizationService {
+  /* eslint-disable-next-line @typescript-eslint/no-empty-function */
+  constructor() {}
+
+  async hasSufficientRole(
+    permittedRoles: Array<string>,
+    appUserToCheck: AppUser
```
godwinpang, line 11: just appUser is fine

src/services/AuthorizationService.ts
```diff
@@ -0,0 +1,17 @@
+import { singleton } from 'tsyringe';
+import { AppUser } from '@Entities';
+
+@singleton()
+export class AuthorizationService {
+  /* eslint-disable-next-line @typescript-eslint/no-empty-function */
+  constructor() {}
```
godwinpang, line 7: o does typescript force you to have one? I thought you could just omit the ctor

src/middlewares/authorization/AuthorizationFactory.ts
```diff
@@ -0,0 +1,65 @@
...
+    const authorizationMiddleware = new AuthorizationMiddleware(
+      authenticationService,
+      authorizationService
+    );
+
+    switch (permissionLevel) {
```
godwinpang, line 28: if you want to clean this up you can have the mapping defined in a const outside the class, then it's just sth like authMiddleware.setPermittedRoles(RoleDef[role]) or sth like that

src/controllers/EventController.ts
```diff
@@ -49,6 +59,7 @@ export class EventController {
   }
 
   @Get('/')
+  @UseBefore(AuthenticationMiddleware, OfficerAuthorizationFactory)
```
godwinpang, line 62: 2 things:

1. Don't think we need AuthenticationMiddleware here because it has authorizationalready
2. The factory is a function that returns a middleware, so I think you just need to call it

great work!

src/middlewares/authorization/AuthorizationFactory.ts
thaigillespie: I'm not sure, I just like destructuring

src/controllers/EventController.ts
thaigillespie: Yeah it's an enum in AppUserEntity

src/services/AuthorizationService.ts
thaigillespie: Probably not, I'll try removing the empty ctor

src/controllers/EventController.ts
godwinpang: cool - was a little confused as I thought it was an entity

src/controllers/EventController.ts
thaigillespie: Should we even have authentication middleware anymore? I mean it's basically already in authorization middleware...

godwinpang: guess not :p

src/constants/ErrResponses.ts
thaigillespie: Should I throw these in auth middleware instead of return next(err_msg)?

src/controllers/EventController.ts
thaigillespie: Maybe I'll just leave it there, whatever use case needs authentication but not authorization can use it (though I know of no such use case rn).

thaigillespie: Actually I just removed the authentication folder lol

PR #53, opened 2020-08-20
No repository as instance variables anymore.

thaigillespie:
They are now obtained ad hoc via calling getRepository(<EntityClass>).
One should keep in mind that regardless of repository not being stored
in instance variables anymore, getRepository should only be called at
the appropriate layer(s) of abstraction.


thaigillespie:
And I did some testing on the events endpoint and everything seems to work fine.

thaigillespie:
Yeah haha I hope this fixes it too :))

PR #54, opened 2020-08-20
Add authentication to OpenAPI schema.

godwinpang:
This allows for client codegen to have authorization features.


src/controllers/EventController.ts
```diff
@@ -42,6 +42,7 @@ export class EventController {
 
   @Post('/')
   @ResponseSchema(EventResponse)
+  @OpenAPI({ security: [{ TokenAuth: [] }] })
```
thaigillespie, line 45: What is TokenAuth?

src/routers/DocsRouter.ts
```diff
@@ -29,6 +32,12 @@ const openAPISpec = routingControllersToSpec(rcMetadataStorage, rcOptions, {
   },
 });
 
+openAPISpec.components.securitySchemes = {};
+openAPISpec.components.securitySchemes[SecuritySchemeName] = {
+  type: 'http',
+  scheme: 'bearer',
```
thaigillespie, line 38: What is 'bearer'?

src/controllers/EventController.ts
godwinpang: it's the security scheme I defined in docs router; time for docs :p

src/routers/DocsRouter.ts
godwinpang: https://swagger.io/docs/specification/authentication/bearer-authentication/
will add comment

PR #77, opened 2020-08-25
WIP: Added endpoint to get user profile.

AneeshJalan:
WIP:
 * Does not currently send back some information related to the profile.


src/payloads/AppUser.ts
```diff
@@ -46,3 +47,23 @@ export class AppUserRolesResponse {
...
+  email: string;
+
+  @IsString()
+  major: string;
+
+  @IsString()
+  graduationYear: string;
+
+  @IsInstance(InductionClass)
+  inductionClass: InductionClass;
```
thaigillespie, line 68: Should make this eager in InductionClass's entity file to fix the problem you had (https://orkhan.gitbook.io/typeorm/docs/eager-and-lazy-relations) 

An example in our codebase (hosts): https://github.com/HKN-UCSD/hkn-server/blob/dev/src/entities/Event.ts

src/mappers/AppUserMapper.ts
```diff
@@ -80,6 +85,23 @@ export class AppUserMapper {
...
+  /**
+   * Converts an AppUser entity to an AppUserProfileResponse payload and returns the newly created
+   * response payload to the caller.
+   *
+   * @param {AppUser} appUser The AppUser entity to be ocnverted to an AppUserProfileResponse payload.
+   * @returns {AppUserProfileResponse} An AppUserProfileResponse instance.
+   */
+  entityToProfileResponse(appUser: AppUser): AppUserProfileResponse {
+    const plainAppUserProfile: Object = classToPlain(appUser);
+    const appUserProfileRespose: AppUserProfileResponse = plainToClass(
```
thaigillespie, line 98: Change Respose to Response.

src/controllers/UserController.ts
```diff
@@ -29,6 +30,19 @@ export class UserController {
...
+
+  @Get('/:userID')
+  @ResponseSchema(AppUserProfileResponse)
+  async getUserProfile(
+    @Param('userID') userID: number,
+    @CurrentUser({ required: true }) appUser: AppUser
+  ): Promise<AppUserProfileResponse | undefined> {
+    const { id } = appUser;
+
+    const appUserObj: AppUser = await this.appUserService.getAppUserById(userID);
```
thaigillespie, line 42: Not sure what behavior we're trying to achieve here, but if it's something along the lines of "members and inductees can only look up their own user profile using their own id, officers can look up user profiles using any user's existing id" then you can probably use the code I wrote for getUserRole as reference.

src/payloads/AppUser.ts
```diff
@@ -1,6 +1,7 @@
-import { IsEnum, IsInt, IsString, IsEmail } from 'class-validator';
+import { IsEnum, IsInt, IsString, IsEmail, IsInstance, ValidateNested } from 'class-validator';
```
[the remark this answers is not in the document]
AneeshJalan: Done

src/payloads/AppUser.ts
AneeshJalan: Cool.. We get induction class in our response now.

src/mappers/AppUserMapper.ts
AneeshJalan: Done

src/controllers/UserController.ts
AneeshJalan: The issue description stated authorization checks to be a non-goal. However, I can work on this if that has changed. @godwinpang @thaigillespie

godwinpang: yeah @thaigillespie it's fine if we address auth in a separate pr now that we have dev and prod separate :)

src/controllers/UserController.ts
```diff
@@ -29,6 +30,19 @@ export class UserController {
...
     return { role: queriedRoleFromId };
   }
+
+  @Get('/:userID')
+  @ResponseSchema(AppUserProfileResponse)
+  async getUserProfile(
+    @Param('userID') userID: number,
+    @CurrentUser({ required: true }) appUser: AppUser
+  ): Promise<AppUserProfileResponse | undefined> {
+    const { id } = appUser;
```
godwinpang, line 40: doesn't eslint yell at you for unused var here?

src/entities/AppUser.ts
```diff
@@ -36,7 +36,8 @@ export class AppUser {
   @Column({ nullable: true })
   graduationYear: string;
 
-  @ManyToOne(() => InductionClass, { nullable: true })
+  @ManyToOne(() => InductionClass, { nullable: true, eager: true })
+  @JoinTable()
```
godwinpang, line 40: why do we need a join table here? it's many to one not many to many

src/payloads/AppUser.ts
```diff
@@ -46,3 +46,23 @@ export class AppUserRolesResponse {
...
+  email: string;
+
+  @IsString()
+  major: string;
+
+  @IsString()
+  graduationYear: string;
+
+  @IsInstance(InductionClass)
+  inductionClass: InductionClass;
```
godwinpang, line 67: eh we can probably just grab the name of the induction class or the quarter or sth

src/entities/AppUser.ts
AneeshJalan: I thought Join table would tell typeorm to get the data from the other table as well instead of just returning the inductionClass value.

godwinpang: doesn't eager do the job? join table creates a new table in the db

src/payloads/AppUser.ts
AneeshJalan: I think most people will remember the quarter in which they inducted. What they won't remember is the name of their induction class.

godwinpang: yeah we can grab the quarter that's a good point

src/entities/AppUser.ts
AneeshJalan: You're right. Removed JoinTable.

src/controllers/UserController.ts
thaigillespie: All the infrastructure for auth and stuff is there as well if you want to try your hands at it @AneeshJalan

src/controllers/UserController.ts
AneeshJalan: When you say yell, do you mean it gives an error or something and doesn't run the program when doing npm run start? Also, IDK why I didn't notice this comment earlier.. I would have changed this as well. I left the ID in because it would be used in auth later so I thought might as well just leave it there.

godwinpang: yeah probably just a warning then - that's fine

src/controllers/UserController.ts
AneeshJalan: This will be done in a separate PR/Issue.

Issue #70, opened 2020-08-23
Startup postgres in a single docker container locally


godwinpang:
tl;dr I want to run `npm run db:start` and `npm run db:stop` to stand up and kill the postgres docker container.
Empty db is *fine*.

you shouldn't need to write a docker file since I *think* we can directly use the base postgres image. Please remember to use the same pg version as we are using on heroku

Non-goal: seeding database

This should be straightforward - tons of guides online

https://docs.docker.com/engine/reference/run/

thaigillespie:
What's our POSTGRES_PASSWORD btw? When I run the command below

`docker run -it --name postgres_container -p <port>:<port> postgres:12.3`

It gives an error and asks me for POSTGRES_PASSWORD

thaigillespie:
Also not really sure if we should run this detached or in foreground https://docs.docker.com/engine/reference/run/#detached-vs-foreground

godwinpang:
foreground is fine - just put anything for postgres_password i think

godwinpang:
u can specify env vars with -e

thaigillespie:
sounds good tysm

thaigillespie:
Still gives me an error even though I specified an env var, it still creates a container tho.
```
Error: Database is uninitialized and superuser password is not specified.
       You must specify POSTGRES_PASSWORD to a non-empty value for the
       superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".

       You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
       connections without a password. This is *not* recommended.

       See PostgreSQL documentation about "trust":
       https://www.postgresql.org/docs/current/auth-trust.html
```

Command: `docker run -it --name postgres_container -p 8080:8080 postgres:12.3 -e POSTGRES_PASSWORD=asfsd`

godwinpang:
Oh put it before you specify the base image maybe

godwinpang:
to clarify - for tests you'll want to run it detached; you should also forward 5432 instead of 8080

godwinpang:
Command: `docker run -it --name postgres_container -p 5432:5432 -e POSTGRES_PASSWORD=asfsd -d postgres:12.3`

PR #87, opened 2020-08-31
Seed to local postgres running in Docker.

### LINKED ISSUE #71: Create test db seed script
~~You can just wipe everything we currently have on the test db and seed it using our old faker scripts (using API instead of firebase obviously) You can then use pg_dump to get a sql script that we can then use to seed the database.~~

See this [link](https://stackoverflow.com/questions/51198817/typeorm-how-to-seed-database). This is probably nicer just so we can run npm run seed:run or sth like that

godwinpang:
closes #71 #70 

Missing some docs for added scripts *and* also docker cmd breaks on windows powershell. People should try out WSL :(


ormconfig.js
```diff
@@ -12,23 +12,40 @@ if (process.env.NODE_ENV === 'test') {
...
-  ssl: true,
-  extra: {
-    ssl: {
-      rejectUnauthorized: false,
+let prodSSLConfig = { ssl: true, extra: { ssl: { rejectUnauthorized: false } } };
+let localSSLConfig = { ssl: false, extra: {} };
+let sslConfig = process.env.NODE_ENV === 'development' ? localSSLConfig : prodSSLConfig;
+
+module.exports = [
+  {
```
thaigillespie, line 20: What is the difference between this and the "seed" object array element?

package.json
```diff
@@ -11,13 +11,24 @@
     "start": "cross-env NODE_ENV=development tsc-watch --onSuccess \"node .\"",
     "prebuild": "npm-run-all clean lint",
     "build": "tsc",
+    "build:no-check": "tsc",
```
thaigillespie, line 14: What does :no-check do to build btw?

src/entities/AppUser.ts
```diff
@@ -36,7 +36,7 @@ export class AppUser {
   @Column({ nullable: true })
   graduationYear: string;
 
-  @ManyToOne(() => InductionClass, { nullable: true, eager: true })
+  @ManyToOne(() => InductionClass, { nullable: true, deferrable: 'INITIALLY DEFERRED' })
```
thaigillespie, line 39: What does the deferrable do here?

package.json
```diff
@@ -11,13 +11,24 @@
...
     "prestart": "npm run build",
     "lint": "eslint . --ext .ts --fix",
     "format": "prettier --write src/**/*.ts",
     "test": "jest",
+    "pretypeorm": "npm-run-all clean build:no-check",
     "typeorm": "node -r module-alias/register ./node_modules/typeorm/cli.js",
-    "typeorm:sync": "node -r module-alias/register ./node_modules/typeorm/cli.js schema:sync",
-    "pretypeorm:sync": "npm-run-all clean build"
+    "typeorm:sync": "npm run typeorm schema:sync",
+    "migration:gen": "npm run typeorm migration:generate --",
```
thaigillespie, line 22: What's the difference between generate and create? Just curious : p

src/migrations/1598831126707-InitialSchema.ts
```diff
@@ -0,0 +1,116 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
thaigillespie, line 1: Please tell me this was autogenerated.

src/migrations/1598829375940-TypeORMMetadata.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
thaigillespie, line 1: This was autogenerated too right????

tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "outDir": "dist",
     "lib": ["ES6"],
     "baseUrl": ".",
+    "incremental": true,
```
thaigillespie, line 14: What does this do?

ormconfig.js
godwinpang: this just specifies 2 separate connections, it's an abuse to use seeds as separate "migrations"

package.json
godwinpang: it just skips prebuild, so no clean no lint. actually now you bring it up we shouldn't skip clean

package.json
godwinpang: gen looks at entities to generate sql that describes the diff, create creates empty

src/entities/AppUser.ts
godwinpang: https://hashrocket.com/blog/posts/deferring-database-constraints#:~:text=In%20PostgreSQL%2C%20constraints%20are%20assumed,behavior%20to%20change%20within%20transactions.

makes seeding easier to deal with since we're inserting events in the same transaction as users

godwinpang: otherwise we'd have to write like 5 seeds inductionclass -> users -> events -> attendance/rsvp

src/migrations/1598829375940-TypeORMMetadata.ts
godwinpang: the structure was generated through migration:create but I handwrote the sql query

src/migrations/1598831126707-InitialSchema.ts
godwinpang: yeah it's just npm run migration:gen

tsconfig.json
godwinpang: Enable incremental compilation by reading/writing information from prior compilations to a file on disk.

godwinpang:
@thaigillespie  this is mainly done, but id prefer not to merge it in until we can verify with docker on windows. otherwise I can take out the docker part and merge. lmk whether or not u have time to test docker

PR #94, opened 2020-09-05
Removed cross-env from migration/seed:run and added to localpg:start (npm commands)

thaigillespie:


godwinpang:
@thaigillespie please rename PR to something more descriptive 

also next time try to pull dev onto dev locally then rebase your branch onto dev - that way your PR is cleaner and will only have the single commit that you have

godwinpang (approved):

PR #95, opened 2020-09-05
Added signup endpoint for AppUsers with non-guest role to signup.

thaigillespie:


src/controllers/AuthController.ts
```diff
@@ -1,175 +1,39 @@
...
-    res.status(403).json({
-      msg: ERR_MSG.USER_NOT_WHITELISTED,
-    });
-    return next();
-  }
-
-  const userDocID: string = userDocDetails.uid;
-
-  let newUser: admin.auth.UserRecord;
+    const appUserWhitelisted = await this.appUserService.getAppUserByEmail(email);
```
godwinpang, line 19: I think we can drop the whitelist here

src/services/FirebaseAdminService.ts
```diff
@@ -0,0 +1,20 @@
+import admin from 'firebase-admin';
+import '@firebase/auth';
+
```
godwinpang, line 3: can we make this a class?

godwinpang: I also think it'd be nice to call this AuthService or AccountService, that way it makes sense if we later don't use firebase anymore

src/services/FirebaseAdminService.ts
```diff
@@ -0,0 +1,20 @@
...
+export const createNewFirebaseUser = async (
+  id: number,
+  email: string,
+  password: string
+): Promise<admin.auth.UserRecord | undefined> => {
```
godwinpang, line 8: To blackbox this service properly I don't think we should return UserRecord. how about just returning the id of the created user?

src/services/AppUserService.ts
```diff
@@ -119,6 +119,14 @@ export class AppUserService {
 
     return false;
   }
+
+  isGuest(appUser: AppUser): boolean {
+    if (appUser.role !== AppUserRole.GUEST) {
```
godwinpang, line 124: `return appUser.role === AppUserRole.GUEST`

godwinpang (changes requested):

src/services/FirebaseAdminService.ts
thaigillespie: I'll call it SignupService. How does that sound?

src/services/FirebaseAdminService.ts
thaigillespie: Yep, sounds good

src/services/AppUserService.ts
thaigillespie: I keep forgetting about this lol...

src/controllers/AuthController.ts
thaigillespie: Renamed to appUserFromEmail

src/services/FirebaseAdminService.ts
godwinpang: I think in general it's better to name things a little more generic - later on we can expose deleteAccount, setVerifyEmail to write scripts and maybe manage accounts. Let's go with accountservice?

thaigillespie: Yep, changed.

src/services/AccountService.ts
```diff
@@ -0,0 +1,33 @@
...
+   * @returns {string} The userID of the newly created Firebase Auth user, if the creation
+   * process succeeds.
+   */
+  async createNewFirebaseUser(
+    id: number,
+    email: string,
+    password: string
+  ): Promise<string | undefined> {
+    try {
+      const newFirebaseUser = await admin.auth().createUser({
```
godwinpang, line 20: later we can inject admin.auth() for testing, but lets merge in for now

godwinpang (approved):

PR #105, opened 2020-09-20
Made AppUserPostRequest payload's major and graduationYear optional.

thaigillespie:
Branch name is wrong bc I was initially going to make a new initialize user endpoint but ended up just making changes to the request payload of create user endpoint : (


thaigillespie:
There are 18 commits because initialize_user_endpoint checks out from dev and master is, for some reason, 16 commits ahead and 16 commits behind dev when they should be equal. Closing this and making a new branch that checks out from master so we don't have 18 commits.

PR #107, opened 2020-09-20
Drop inheritance for Event payloads.

thaigillespie:
This was the only one payload file left that had inheritance.


godwinpang (approved):

godwinpang:
Could you rename PR to Drop inheritance vs Dropped inheritance to adhere to standard git naming conventions? e.g. Merge commits are merge instead of merged

PR #110, opened 2020-09-23
Checkoff API implemented and manually tested and getAllAppUsers bug fix.

thaigillespie:
Probably should have fixed that bug in a separate PR but it's such a minor fix : /


src/controllers/EventController.ts
```diff
@@ -105,8 +117,46 @@ export class EventController {
...
+    @Param('eventID') eventID: number,
+    @Body() attendanceCheckOffRequest: AttendanceCheckOffRequest
+  ): Promise<AttendanceResponse | undefined> {
+    const { attendeeId, officerId, ...otherValues } = attendanceCheckOffRequest;
+
+    if (officerId == undefined) {
+      return undefined;
+    }
+
+    const [event, attendee, officer] = await Promise.all([
```
godwinpang, line 150: we don't actually need to query the objects, we should be able to save just with the id
i think i did it with the events where we had hosts as the PKpayload

src/services/AttendanceService.ts
```diff
@@ -1,8 +1,60 @@
...
+  }
+
+  /**
+   * Gets all attendances of a specified that are filtered based on passed in query parameters.
+   *
+   * @param {Event} event Event to get attendances from.
+   * @param {MultipleAttendanceQuery} multipleAttendanceQuery Query parameters to filter attendances.
+   * @returns {Promise} Array of attendances from the specified event.
+   */
+  async getAllAttendancesOfEvent(
```
godwinpang, line 42: getAllEventAttendances

godwinpang (changes requested):

src/controllers/EventController.ts
thaigillespie: I think this can only be done for officer column, since event and attendee columns are primary columns @godwinpang

thaigillespie: nvm I made it work

godwinpang:
:) would be nice to document resolution

thaigillespie:
Hmm I guess here is fine?

Any way, things weren’t working because I passed in the raw id instead of
an id wrapped in a JS object. After I made that change everything works
fine. I thought I was doing what Godwin did with AppUserPKRequest but nope.

src/services/EventService.ts
```diff
@@ -54,6 +55,24 @@ export class EventService {
     return event ? eventRepository.remove(event) : undefined;
   }
 
+  async getAttendancesFromEvent(
```
godwinpang, line 58: maybe getEventAttendances? usually don't see from in method names

src/payloads/Attendance.ts
```diff
@@ -1,22 +1,62 @@
...
+export class MultipleAttendanceResponse {
+  @ValidateNested({ each: true })
+  @Type(() => AttendanceResponse)
+  attendances: AttendanceResponse[];
+}
+
+export class MultipleAttendanceQuery {
+  @IsBoolean()
+  @IsOptional()
+  unchecked: boolean;
```
godwinpang, line 57: what's unchecked?

src/payloads/Attendance.ts
```diff
@@ -1,22 +1,62 @@
...
+  IsInt,
+} from 'class-validator';
+import { Type } from 'class-transformer';
+
 import { AppUserEventResponse } from './AppUser';
 import { EventAttendanceResponse } from './Event';
 
+export class AttendanceCheckOffRequest {
+  @IsInt()
+  readonly attendeeId: number;
```
godwinpang, line 16: we need to commit to either Id or ID, i think there's both in the codebase rn

godwinpang (changes requested):

src/payloads/Attendance.ts
thaigillespie: Whether an Attendance is checked off or not

thaigillespie: So if unchecked is true, then we query only the Attendances that are not checked off

src/payloads/Attendance.ts
thaigillespie: Yeah that's true, I personally prefer Id but ID might be used in the codebase more...

thaigillespie: Okay so Id has 53 results in 12 files of hkn-server and ID has 99 results in 17 files of hkn-server. If we do change Id -> ID, I think we might have to be careful because the stuff from typeorm uses Id instead of ID

PR #113, opened 2020-09-25
Fix typeorm preload merging issue with many-to-many relations

thaigillespie:
Manually updated the hosts property in EventMapper's
requestToExistingEntity.

### DIFF
--- a/src/mappers/EventMapper.ts
+++ b/src/mappers/EventMapper.ts
@@ -1,6 +1,6 @@
 import { EventRequest, EventResponse } from '@Payloads';
 import { AppUserService, AppUserServiceImpl } from '@Services';
-import { Event } from '@Entities';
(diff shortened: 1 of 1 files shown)


src/mappers/EventMapper.ts
```diff
@@ -36,6 +40,10 @@ export class EventMapper {
     // preload ignores empty arrays and loads arrays anyways.
     if (eventObj.hosts.length == 0) {
       event.hosts = [];
+    } else {
+      event.hosts = await Promise.all(
+        hosts.map(host => this.appUserService.getAppUserById(host.id))
```
godwinpang, line 45: just put in ID

[{id:userID}]

godwinpang (changes requested):

godwinpang (approved):

PR #114, opened 2020-09-27
Add docs for git and SuperTest

thaigillespie:


guides/testing/supertest.md
```diff
@@ -0,0 +1,7 @@
+# Supertest
+
```
godwinpang, line 2: probably good to explain that we use supertest for e2e testing :)

godwinpang (changes requested):

guides/testing/supertest.md
thaigillespie: Ah yeah that's true, let me add that

thaigillespie: Added @godwinpang

godwinpang:
@frankyyzy can you take a look at this PR and approve it if the docs look good?

godwinpang (approved):

PR #118, opened 2020-09-30
Change dotenv from a prod dependency to a dev dependency

thaigillespie:


ormconfig.js
```diff
@@ -20,8 +18,8 @@ module.exports = [
   {
     type: 'postgres',
     url: process.env.DATABASE_URL,
-    ssl: sslConfig.ssl,
-    extra: sslConfig.extra,
+    ssl: prodSSLConfig.ssl,
```
godwinpang, line 21: why is this being done? are we not doing sslConfig based on node_env anymore?

godwinpang (changes requested):

ormconfig.js
thaigillespie: Ah crap, this is because I was testing on dev db :)))))

ericke8 (approved):
removing dotenv from everything + script change looks good

PR #129, opened 2020-10-21
Fix issue of incorrect object spreading with query object for getAllEventAttendances

thaigillespie:


src/services/AttendanceService.ts
```diff
@@ -17,7 +17,8 @@ export class AttendanceService {
    */
   private buildMultipleAttendanceQuery(
     event: Event,
-    multipleAttendanceQuery: MultipleAttendanceQuery
+    multipleAttendanceQuery: MultipleAttendanceQuery,
+    cacheOn: boolean
```
godwinpang, line 21: withCache

godwinpang (changes requested):

godwinpang (approved):

PR #135, opened 2020-10-24
Add endpoints for getting member + inductee points

godwinpang:
RN this gives a 404 when you query for someone who doesn't have points which probably isn't the best but uh it works so idc for now :D


src/entities/Attendance.ts
```diff
@@ -25,7 +25,7 @@ export class Attendance {
   @Column('timestamp', { nullable: true })
   endTime?: Date;
 
-  @Column({ nullable: true })
+  @Column({ type: 'float4', nullable: true })
```
thaigillespie, line 28: This gives 6 decimal degits precision. Any reason why you picked this? Just curious.

src/controllers/UserController.ts
```diff
@@ -117,6 +119,41 @@ export class UserController {
...
+    @CurrentUser({ required: true }) requestAppUser: AppUser
+  ): Promise<AppUserInducteePointsResponse | undefined> {
+    if (this.appUserService.isInvalidNonOfficerAccess(requestAppUser, userID)) {
+      throw new ForbiddenError();
+    }
+
+    const points = await this.appUserService.getInducteePoints(userID);
+
+    // I don't like this but I'll take it for now - Godwin Oct 24 2020
+    return points;
```
thaigillespie, line 138: Perhaps renaming it to pointsObj or smth is better? There is a bit more information that just the points themselves.

src/controllers/UserController.ts
```diff
@@ -117,6 +119,41 @@ export class UserController {
...
+    @Param('userID') userID: number,
+    @CurrentUser({ required: true }) requestAppUser: AppUser
+  ): Promise<AppUserMemberPointsResponse | undefined> {
+    if (this.appUserService.isInvalidNonOfficerAccess(requestAppUser, userID)) {
+      throw new ForbiddenError();
+    }
+
+    const points = await this.appUserService.getMemberPoints(userID);
+
+    return points;
```
thaigillespie, line 155: Same as above for this

thaigillespie (changes requested):
Lgtm but had some small things I wanted to bring up.

src/entities/Attendance.ts
godwinpang: it works :)

src/controllers/UserController.ts
godwinpang: idr why i did this don't ask :|

godwinpang: I think points is fine for now... it's mainly just points right? idt pointsObj is strictly a better name, maybe induction requirements might be better lol

thaigillespie: Yeah sounds good, I'll approve this then

thaigillespie (approved):
lgtm

PR #136, opened 2020-10-31
Add query params to GET /api/events

thaigillespie:


src/services/EventService.ts
```diff
@@ -2,12 +2,43 @@ import { Event, AppUser, Attendance, RSVP } from '@Entities';
...
+    const query: FindManyOptions<Event> = {};
+    const whereArr = [];
+
+    if (canShowPending) {
+      if (pending) {
+        whereArr.push({ status: 'pending' });
+      }
+    } else {
+      if ((ready && complete) || (!ready && !complete)) {
+        whereArr.push({ status: Not('pending') });
```
godwinpang, line 25: whats this

godwinpang (changes requested):

src/services/EventService.ts
thaigillespie: This is because Not('pending') queries include 'ready' and 'complete' statuses (so it gets anything that is not 'pending'). So let's say a user is not permitted to see pending events, but say they only want to see events that are ready but not events that are complete. Then, without that check, Not('pending') will erroneously return both events that are ready and events that are complete, even though we only wanted events that are ready. By having that check, we are making sure that Not('pending') = 'ready' + 'complete' query is made only when we have both ready and complete = true in query params or when we don't have either of them, since no query params = show every event but for some users we don't want them to see pending so they'll only be seeing ready and complete with this.

thaigillespie:
Note that we still need a mechanism for officers to change the status of an event on the frontend, since we don't have that yet.

src/services/EventService.ts
```diff
@@ -2,12 +2,43 @@ import { Event, AppUser, Attendance, RSVP } from '@Entities';
...
-import { MultipleAttendanceQuery } from '@Payloads';
+import { getRepository, FindManyOptions, Not } from 'typeorm';
+import { MultipleAttendanceQuery, MultipleEventQuery } from '@Payloads';
 
 export class EventService {
   constructor(private attendanceService: AttendanceService, private rsvpService: RSVPService) {}
 
+  private buildMultipleEventQuery(
+    multipleEventQuery: MultipleEventQuery,
+    canShowPending: boolean
```
godwinpang, line 13: what about isOfficer here? then the logic for who should see pending is a little more clear - if you are an officer then you will be able to see xyz.

Another option that might be a little nicer would be to actually pass in the entire AppUser and do the logic of who can see what within the EventService for better encapsulation(not 100% sure on this one kinda up to you)

src/services/EventService.ts
```diff
@@ -2,12 +2,43 @@ import { Event, AppUser, Attendance, RSVP } from '@Entities';
...
+  ): FindManyOptions<Event> {
+    const { pending, ready, complete } = multipleEventQuery;
+    const query: FindManyOptions<Event> = {};
+    const whereArr = [];
+
+    if (canShowPending) {
+      if (pending) {
+        whereArr.push({ status: 'pending' });
```
godwinpang, line 21: for the 'pending' strings can we reuse the constants from where we defined them in the Event Entity file?

godwinpang (commented):
sorry for leaving this open for so long - lmk what you think about the comments below and we can close this up :)

src/services/EventService.ts
thaigillespie: I'm not sure if isOfficer is a good name, since there is no indication of what being an officer querying for multiple events entail. If we do use isOfficer, it is probably best to document what being an officer means in this context (what it allows an officer to see, as opposed to a non-officer member).

I agree with your second point, I think it would be nice to make the query builder function as generic as possible so too much logic is not being jammed into it

thaigillespie: To add on to your second point, I think it's best then to maybe not have the query builder function for GET multiple events. Can just jam the logic into the service function for GET multiple events or smth

godwinpang: up to you - lmk what you decide on

thaigillespie: @godwinpang new commit pushed

godwinpang (approved):

PR #137, opened 2020-10-31
Add guide for adding new entity.

godwinpang:


thaigillespie (approved):

frankyyzy:
I think "npm run localpg:start" will also seed the database, so after modification of the entity files, "npm run migration:run" needs to be run first, otherwise data won't be inserted into the database in the correct format. For example, if I added a non-null column for induction class, and run "npm run localpg:start" first, I'll get an error saying that column is null.

PR #139, opened 2020-11-08
Initial setup of getting user availabilities for interview

AimeeShao:
I'm thinking what I've written should work... but I have no experience...

I made an Availability table/entity, to store each individual (date : jsonb of start and end time).
Notes:

1. idk if the `@Column({ type: 'jsonb'})` works. This is what i got from reading multiple threads on "typeorm/postgres column array of jsonb"

- I thought making jsonb column (if possible) would look nicer and simpler in code compared to creating additional entities of AvailabilityDates which consists of array of entity TimeRanges or something like that

2. AppUsers currently has `@ManyToOne(() => InductionClass,`. According to what I learned about OneToMany and ManyToOne, this means users can only have one induction class, but if we want them to have multiple, i can include that in my changes.

- And if they do have multiple, I probably need to make another Entity (Availabilities) to encapsulate Availability[] (in which case i could also move the OneToMany and ManyToOne relationship between user and availabilities out, and just make it a column relationship which makes more sense...)

- Let me know if i should do the above, it seems to make the most sense, although i am initially unsure about creating two entities.


src/entities/AppUser.ts
```diff
@@ -45,4 +46,11 @@ export class AppUser {
     default: AppUserRole.GUEST,
   })
   role: string;
+
+  @OneToMany(
+    () => Availability,
+    availability => availability.user,
+    { nullable: true, deferrable: 'INITIALLY DEFERRED' }
+  )
+  availabilities: Availability[];
```
godwinpang, line 55: as mentioned in discord - we don't need an additional `Availability` table

src/payloads/AppUser.ts
```diff
@@ -191,3 +191,15 @@ export class MultipleUserNameResponse {
   @Type(() => AppUserNameResponse)
   users: AppUserNameResponse[];
 }
+
+export class InductionClassQuery {
+  @IsString()
+  @IsOptional()
+  induction_class: string;
```
godwinpang, line 198: camel case

godwinpang (commented):
Good work @AimeeShao! everything looks right - just some minor comments base on the followups we had on Discord

how did you make the PR a draft? :O

src/entities/AppUser.ts
```diff
@@ -45,4 +45,14 @@ export class AppUser {
...
   })
   role: string;
+
+  @Column({ nullable: true, type: 'json' })
+  availabilities: Availabilities;
+}
+
+export type Availabilities = TimeRange[];
+
+interface TimeRange {
```
thaigillespie, line 55: Put interface definition to the top of the file to conform to convention of our codebase

thaigillespie (approved):
Lgtm, left a little comment, you can fix it right away if you want, but leaving this for another PR is fine as well

thaigillespie (approved):

PR #140, opened 2020-11-12
Change startTime and endTime of Attendance from type Date to type string

thaigillespie:
This includes startTime, endTime type changes in src/entities and src/payloads


thaigillespie:
Thought I had to run a migration for this but turns out when we originally added startTime and endTime and points stuff to Attendance, the generated SQL already stored them as timestamps so I didn't have to do any migration work.

godwinpang (approved):

Issue #130, opened 2020-10-24
add endpoint: GET /inductionclass/{id}/interviewdates 


frankyyzy:

frankyyzy:
use induction quarter as id

PR #155, opened 2020-11-20
Interview dates endpoint

frankyyzy:
#130


src/entities/InductionClass.ts
```diff
@@ -22,4 +23,10 @@ export class InductionClass {
   // Starting date of the induction cycle for said class
   @Column('date')
   endDate: string;
+
+  @Column({
+    type: 'simple-json',
```
godwinpang, line 28: we should use json instead of simple-json here to rely on the native json column type

godwinpang: might have to do more magic with getting types to work nicely idk

src/entities/InductionClass.ts
```diff
@@ -7,7 +7,8 @@ import { Entity, PrimaryColumn, Column } from 'typeorm';
 @Entity()
 export class InductionClass {
   // TODO add validation for what strings are allowed.
-  // e.g. FA20
+
+  // e.g. fa20 (lower case to match string in url)
```
godwinpang, line 11: we shouldn't be changing this, existing data is already in upper case

godwinpang (changes requested):

src/payloads/InterviewDates.ts
```diff
@@ -0,0 +1,6 @@
+import { IsArray } from 'class-validator';
+
+export class InterviewDatesResponse {
+  @IsArray()
+  interview: Date[];
```
godwinpang, line 5: We should try to have payloads as strings - Date gets converted into ISO string in json anyways
Better to be explicit here

godwinpang (changes requested):

thaigillespie:
Gonna just merge this now

PR #161, opened 2020-12-22
Make separate endpoints for affiliate event signin and RSVP

thaigillespie:
Create separate endpoints for affiliate event signin and RSVP so that we don't have to do that ugly thing on frontend where we have to add filler inputs to the POST request body of the original endpoints for affiliate signin/rsvp so it works (really hacky). I also made it so that affiliates won't be able to use the non-affiliate event signin/rsvp endpoints (they will get 403 - "Forbidden" - error code).

### DIFF
--- a/src/controllers/EventController.ts
+++ b/src/controllers/EventController.ts
@@ -8,6 +8,8 @@ import {
   UseBefore,
   CurrentUser,
   QueryParams,
+  ForbiddenError,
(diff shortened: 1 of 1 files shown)


godwinpang (approved):

thaigillespie:
Gonna merge it now

thaigillespie:
Will do integration later

PR #163, opened 2020-12-22
Add docker-compose and Dockerfile.

godwinpang:
This allows for setting up the whole backend stack locally via a single
docker-compose up command. Paves the way for having a monorepo and
setting up backend + frontend stack in a single command, allowing for
e2e tests.


godwinpang:
Note that if running `docker-compose up` the db url will have to have db instead of localhost as the hostname.

godwinpang:
dev images contain node_modules and npm i takes ~20 minutes to run
abandoning for now

PR #166, opened 2020-12-29
Add optional status field to EventRequest payload for updateEvent

thaigillespie:


thaigillespie:
Just gonna squash merge this since it's a small change that I tested this somewhat on my computer (and it looks like it works).

PR #169, opened 2021-01-25
Add endpoint to get RSVPs, add extra functionality to query params of getEventAttendances()

thaigillespie:
I decided to keep everything related to attendance and rsvp in EventController for now

### DIFF
--- a/src/services/RSVPService.ts
+++ b/src/services/RSVPService.ts
@@ -36,7 +36,7 @@ export class RSVPService {
     return newRSVP;
   }
 
-  async getAllEventRSVPs(event: Event): Promise<RSVP[]> {
+  async getEventRSVPs(event: Event): Promise<RSVP[]> {
     const rsvpRepository = getRepository(RSVP);
     const query = this.buildMultipleRSVPQuery(event, true);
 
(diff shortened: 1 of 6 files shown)


godwinpang (approved):
👍🏻  but find someone else to look at it if u can

src/services/RSVPService.ts
```diff
@@ -22,6 +35,13 @@ export class RSVPService {
 
     return newRSVP;
   }
+
+  async getAllEventRSVPs(event: Event): Promise<RSVP[]> {
```
godwinpang, line 39: maybe rename to getEventRSVPs? right now it sounds like getting all the events's RSVPs

godwinpang: also jw here (probably remnant from before), why is the param an event instead of eventId?
no problem with it being event im just curious

godwinpang (approved):

src/services/RSVPService.ts
thaigillespie: For your second comment, I think it's because I do event existence check first, so I figured I might as well just pass in the event object itself since I have it (I think this was my thought process when I first worked on attendance/rsvp stuff, but man it's been a while so idk)

thaigillespie: for the first question maybe idk I'm lowkey a bit lazy lol

godwinpang (approved):

PR #170, opened 2021-01-29
S3 eco

ericke8:
basic functionalities for storage

includes:
StorageController - privileged access to storage by officer only
UserController - added endpoints for resume upload and download

StorageService - storage adapter for S3
LocalStorageService - storage adapter for local fs
ResumeService - service specifically for storing/getting resume


src/controllers/StorageController.ts
```diff
@@ -0,0 +1,88 @@
...
+  UploadedFile,
+  BadRequestError,
+  UseBefore,
+} from 'routing-controllers';
+import { AppUserService, AppUserServiceImpl, StorageService, StorageServiceImpl } from '@Services';
+import multer from 'multer';
+import { OfficerAuthMiddleware } from '@Middlewares';
+import { OpenAPI } from 'routing-controllers-openapi';
+
+const fileFilter = (req: Express.Request, file: Express.Multer.File, cb: Function) => {
```
godwinpang, line 15: would be nice to have some documentation here linked to multer - initial reaction to this might just be why not promises

src/controllers/StorageController.ts
```diff
@@ -0,0 +1,88 @@
...
+  ) {
+    cb(null, true);
+  } else {
+    console.log('Invalid file type');
+    cb(null, false);
+  }
+};
+
+const fileUploadOptions = {
+  storage: multer.memoryStorage(),
```
godwinpang, line 29: could we add a short comment on this? sounds scary that we're using memoryStorage (but probably not form data)

src/services/StorageService.ts
```diff
@@ -0,0 +1,97 @@
...
+      Body: file.buffer,
+      ContentDisposition: `attachment; filename="${fileName}${file.originalname.substring(
+        file.originalname.lastIndexOf('.')
+      )}"`,
+      ContentType: `${file.mimetype}`,
+    };
+
+    return s3
+      .upload(params)
+      .promise()
```
godwinpang, line 28: is it possible to switch to await here?

src/services/StorageService.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (options) {
+      if ('appendFileName' in options) {
+        fileNameKey = `${fileName}_${options['appendFileName']}`;
+      }
+    }
+    const params = {
+      Bucket: process.env.BUCKET_NAME,
+      Key: fileNameKey,
+      Body: file.buffer,
+      ContentDisposition: `attachment; filename="${fileName}${file.originalname.substring(
```
godwinpang, line 20: could we add a comment for this? looks like it's trying to get the filename without the extension; any specific reason for that?

src/services/StorageService.ts
```diff
@@ -0,0 +1,97 @@
...
+    options?: Object
+  ): Promise<string | null> {
+    let fileNameKey = fileName;
+    if (options) {
+      if ('appendFileName' in options) {
+        fileNameKey = `${fileName}_${options['appendFileName']}`;
+      }
+    }
+    const params = {
+      Bucket: process.env.BUCKET_NAME,
```
godwinpang, line 17: as mentioned above we should try to keep process.env only in config.ts; will update documentation for this sorry!

src/services/StorageService.ts
```diff
@@ -0,0 +1,97 @@
...
+        file.originalname.lastIndexOf('.')
+      )}"`,
+      ContentType: `${file.mimetype}`,
+    };
+
+    return s3
+      .upload(params)
+      .promise()
+      .then(
+        function(data: AWS.S3.ManagedUpload.SendData) {
```
godwinpang, line 30: what do you think about doing fat arrow function; mainly for consistency

src/services/StorageService.ts
```diff
@@ -0,0 +1,97 @@
...
+          console.log(`Could not retrieve file: ${e.message}`);
+          return null;
+        });
+    } catch (e) {
+      console.log(`Could not retrieve file from S3: ${e.message}`);
+      return null;
+    }
+  }
+
+  async getFileIndex(): Promise<Array<string> | null> {
```
godwinpang, line 67: what does getFileIndex do?

godwinpang (changes requested):
just some general comments
I assume StorageController is for demo purposes?
Not sure what we should do about console.logs will have to see what @frankyyzy comes up with :)

Will localfs break cross platform with "/" vs "\"? (another reason to just get into docker...)

src/services/StorageService.ts
```diff
@@ -0,0 +1,97 @@
+import AWS from 'aws-sdk';
+
+const s3 = new AWS.S3({ region: process.env.BUCKET_REGION });
+export class StorageService {
+  async uploadFile(
+    fileName: string,
+    file: Express.Multer.File,
+    options?: Object
```
godwinpang, line 8: would be nice to have an exported interface for options type instead of Object

src/services/StorageService.ts
ericke8: do we want to be handling errors in the services or on the controllers level?
we should await these calls where we want to use a try catch to output errors

src/app.ts
```diff
@@ -16,6 +16,13 @@ import morgan from 'morgan';
 
 import { checkCurrentUserToken } from './decorators';
 
+import AWS from 'aws-sdk';
```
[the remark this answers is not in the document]
ericke8: done

src/controllers/StorageController.ts
ericke8: done

src/controllers/StorageController.ts
ericke8: done

src/services/StorageService.ts
ericke8: done

src/services/StorageService.ts
ericke8: done

src/services/StorageService.ts
ericke8: done

src/services/StorageService.ts
ericke8: it gets a list of all the files currently in the bucket

src/services/StorageService.ts
ericke8: done

src/services/StorageService.ts
```diff
@@ -0,0 +1,122 @@
...
+      const params = {
+        Bucket: config.awsConfig.bucketName,
+        Key: fileNameKey,
+        Body: file.buffer,
+        ContentDisposition: `attachment; filename="${fileName}${file.originalname.substring(
+          file.originalname.lastIndexOf('.')
+        )}"`,
+        ContentType: `${file.mimetype}`,
+      };
+
```
godwinpang, line 50: shrink try block

src/controllers/UserController.ts
```diff
@@ -34,13 +40,33 @@ import {
 import { AppUserMapper, AppUserMapperImpl } from '@Mappers';
 import { InducteeAuthMiddleware, MemberAuthMiddleware, OfficerAuthMiddleware } from '@Middlewares';
 import { formatISO } from 'date-fns';
+import multer from 'multer';
+
+const fileFilter = (req: Express.Request, file: Express.Multer.File, cb: Function) => {
```
godwinpang, line 45: move to resume service

src/controllers/StorageController.ts
```diff
@@ -0,0 +1,108 @@
+import {
```
godwinpang, line 1: move this file to docs

godwinpang (approved):

godwinpang (approved):
lgtm!

src/services/ResumeService.ts
```diff
@@ -0,0 +1,58 @@
...
+const fileFilter = (req: Express.Request, file: Express.Multer.File, cb: Function): void => {
+  if (file.mimetype.includes('pdf')) {
+    cb(null, true);
+  } else {
+    console.log('Invalid file type');
+    cb(null, false);
+  }
+};
+
+export const resumeFileUploadOptions = {
```
AneeshJalan, line 16: Is there a better spot to place this? It's position here makes it feel like it will be used in the uploadResume function to filter the file but it is actually instead used with the react-controller/multer set up in UserController.

AneeshJalan, line 16: Is there a better spot to place this? It's position here makes it feel like it will be used in the uploadResume function to filter the file but it is actually instead used with the react-controller/multer set up in UserController.

ericke8: Technically, it is used with the uploadResume function, just one level above on the controller for uploadResume.
It's here since the filter is specific to only the resume upload/download, and this file tries to include all the things to do with with resume upload/download.  It wouldn't make sense to be in UserController, since this options object only applies to resume storage.

PR #202, opened 2021-03-01
Logging mono

frankyyzy:
In this branch logging is setup, so that developers can import a logger as"import logger from '@Logger';" and use the looger as "logger.info({
      message: 'msg',
      customTags: 'tag1',
    });"

Some examples are in TeapotController.ts and EventController.ts

For local dev, "MAX_LOG_LEVEL=debug" would need to be added to the .env file to show all logs and can be changed to show only certain levels of log information. Environment variables also need to be set up on Heroku to configure the max level of logs to be displayed.


backend/jest.config.js
```diff
@@ -15,5 +15,6 @@ module.exports = {
     '@Repositories': '<rootDir>/src/repositories',
     '@Middlewares': '<rootDir>/src/middlewares',
     '@Config': '<rootDir>/src/config.ts',
+    '@Logger': '<rootDir>/src/logging.ts',
```
thaigillespie, line 18: Perhaps putting this in a separte folder is better? Just to conform to other folders. Up to you on this one tbh

backend/src/logging.ts
```diff
@@ -1,8 +1,9 @@
 import winston from 'winston';
 
+const logLevels = { error: 0, warn: 1, info: 2, debug: 3 };
+
 const logger = winston.createLogger({
-  level: process.env.WINSTON_LOG_LEVEL,
-  transports: [new winston.transports.Console()],
+  levels: logLevels,
+  transports: [new winston.transports.Console({ level: process.env.MAX_LOG_LEVEL })],
```
thaigillespie, line 7: I noticed that you added MAX_LOG_LEVEL in the destructuring assignment of process.env in config.ts so I think it's better if you add MAX_LOG_LEVEL to the config object in config.ts then import config in this file.

thaigillespie (changes requested):
Lgtm, just a couple of things that I think is good to change

backend/src/logging.ts
frankyyzy: Good idea. Done!

backend/jest.config.js
frankyyzy: Personally, I think it's fine since it's just one file. @godwinpang  what do you think?

godwinpang: yeah I think it's actually better as its own file since there really shouldn't be too much for logging, if our logging setup gets complicated enough then we can move it into its folder and still be protected by the @Logger abstraction

backend/src/controllers/EventController.ts
```diff
@@ -95,6 +96,10 @@ export class EventController {
   @Get('/:eventID')
   @ResponseSchema(EventResponse)
   async getEvent(@Param('eventID') eventID: number): Promise<EventResponse> {
+    logger.info({
+      message: 'getting event ' + eventID,
+      customTags: 'event',
```
godwinpang, line 101: could we take this out for now before we come up with a standardized way of approaching tags?

backend/guides/logging/how_to_add_log.md
```diff
@@ -0,0 +1,35 @@
...
+import logger from '@Logger';
+```
+
+Then to use it, do
+
+```javascript
+logger.log({
+  level: 'debug',
+  message: 'teapot info',
+  customTags: 'testing',
```
godwinpang, line 23: customTags suggests that we can add multiple tags to a log message - would it be possible to look into that as well?

backend/src/controllers/TeapotController.ts
```diff
@@ -1,10 +1,21 @@
...
 
 @JsonController('/api/teapot')
 export class TeapotController {
   @Get('/')
   @HttpCode(418)
   async getEvent(): Promise<string> {
+    logger.log({
+      level: 'info',
+      message: 'teapot info',
+      customTags: 'testing',
```
godwinpang, line 12: see above

backend/src/controllers/TeapotController.ts
```diff
@@ -1,10 +1,21 @@
...
+    });
+    logger.log({
+      level: 'debug',
+      message: 'teapot info',
+      customTags: 'testing',
```
godwinpang, line 17: see above

godwinpang (changes requested):
thanks @frankyyzy! just one small thing about removing the log tags/statements, and we can look into the comment about multiple tags later

backend/guides/logging/how_to_add_log.md
frankyyzy: Yes it's possible to add multiple tags. Details are in the docs.

backend/src/controllers/TeapotController.ts
frankyyzy: done.

backend/src/controllers/TeapotController.ts
frankyyzy: done

backend/src/controllers/EventController.ts
frankyyzy: done.

thaigillespie:
Should we close this then @godwinpang since you merged the other branch with Frank's commits already?

godwinpang:
see #213

PR #208, opened 2021-03-29
feature: add status colors

### LINKED ISSUE #188: Event Color Coding + Statuses on list view
Events should be color coded on the calendar, and their statuses should be shown on the list view.

yuw006:
closes #188

### DIFF
--- a/frontend/src/pages/CalendarPage/components/Calendar/index.js
+++ b/frontend/src/pages/CalendarPage/components/Calendar/index.js
@@ -28,11 +28,6 @@ const resources = [
 
 const mainResource = 'status';
 
-resources.propTypes = {
-  fieldName: PropTypes.string.isRequired,
-  instances: PropTypes.object.isRequired,
-};
-
 const AppointmentWithClick = handleClick => props => (
   <Appointments.Appointment
     {...props}
--- a/frontend/src/pages/CalendarPage/components/Calendar/index.js
+++ b/frontend/src/pages/CalendarPage/components/Calendar/index.js
@@ -10,6 +10,7 @@ import {
   ViewSwitcher,
   TodayButton,
   DateNavigator,
+  Resources,
 } from '@devexpress/dx-react-scheduler-material-ui';
 import { ViewState } from '@devexpress/dx-react-scheduler';
 
@@ -25,7 +26,7 @@ AppointmentWithClick.propTypes = {
   data: PropTypes.object.isRequired,
 };
 
-export default function Calendar({ events, handleEventClick }) {
+export default function Calendar({ events, handleEventClick, resources }) {
   return (
     <Scheduler data={events} firstDayOfWeek={1}>
       <ViewState defaultCurrentViewName='Week' />
@@ -41,6 +42,7 @@ export default function Calendar({ events, handleEventClick }) {
           handleEventClick(appointment)
         )}
       />
+      <Resources data={resources} mainResourceName='status' />
     </Scheduler>
   );
 }
@@ -54,4 +56,5 @@ Calendar.propTypes = {
     })
   ).isRequired,
   handleEventClick: PropTypes.func.isRequired,
+  resources: PropTypes.arrayOf(PropTypes.object.isRequired).isRequired,
 };
--- a/frontend/src/pages/CalendarPage/components/Calendar/index.js
+++ b/frontend/src/pages/CalendarPage/components/Calendar/index.js
@@ -14,6 +14,25 @@ import {
 } from '@devexpress/dx-react-scheduler-material-ui';
 import { ViewState } from '@devexpress/dx-react-scheduler';
 
+// color of events based on status
+const resources = [
+  {
+    fieldName: 'status',
+    instances: [
+      { id: 'complete', color: '#b3cce6' },
+      { id: 'pending', color: '#e6b3cc' },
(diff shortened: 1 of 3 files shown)


yuw006:
I used [Resources](https://devexpress.github.io/devextreme-reactive/react/scheduler/docs/reference/resources/)

frontend/src/pages/CalendarPage/index.js
```diff
@@ -123,6 +123,18 @@ class CalendarPage extends React.Component {
...
 
+    // color of events based on status
+    const resources = [
+      {
+        fieldName: 'status',
+        title: 'Status',
+        instances: [
+          { id: 'complete', text: 'Laser Tag', color: 'cadetBlue' },
+          { id: 'pending', text: 'Test Event', color: 'plum' },
+          { id: 'ready', text: 'Test Event', color: 'tomato' },
```
thaigillespie, line 134: Since text is an optional field (https://devexpress.github.io/devextreme-reactive/react/scheduler/docs/reference/resources/#resourceinstance) you don't have to include it here in these JS objects. If you do want to include text, I think doing something like `text: 'Readied Event'` and `'text: Pending Event'` would be a lot better.

thaigillespie (changes requested):
Overall lgtm. I haven't used Resources so I'm not the best judge at whether or not you're using it well here (but the code looks right and your screenshots indicate that it works so it's good to me). Just had one small comment and I'll approve when you're done resolving it. 

Also, if possible, can you upload some photos of how the status colors of the events look with the whole calendar page in the description of your PR?

frontend/src/pages/CalendarPage/index.js
```diff
@@ -138,36 +150,41 @@ class CalendarPage extends React.Component {
...
-              control={
-                <Checkbox
-                  name='complete'
-                  onChange={() => this.handleCompleteChange()}
-                  checked={complete}
-                />
-              }
-              label='Complete'
-            />
+            {view === 'calendar' &&
```
godwinpang, line 153: For this block could just have 1 ternary statement? view==='calendar' can just control all of the 3 FormControlLabels.

frontend/src/pages/CalendarPage/index.js
```diff
@@ -123,6 +123,18 @@ class CalendarPage extends React.Component {
     } = this.state;
     const { classes, history } = this.props;
 
+    // color of events based on status
```
godwinpang, line 126: What do you think about having this within the Calendar component instead of being in the CalendarPage? I think there's merit in doing so because I don't see CalendarPage as being responsible for controlling *how* the Calendar component should color the events.

Specifically, resources is of an object type that's directly from the dx-scheduler component. Having to define resources outside of the Calendar component tightly couples our CalendarPage to our *implementation* of the Calendar component. For example, if we wanted to switch away from using dx-scheduler in the future, we'd have to change both the CalendarPage and the Calendar component - this means that the abstraction that the Calendar component provides is a "leaky" one.

frontend/src/pages/CalendarPage/index.js
godwinpang: +1 here - I agree with Thai; probably would be nicer to leave out text, not sure what its purpose is here as of now

godwinpang (changes requested):

yuw006:
I fixed the the resources field, and placement of the variable. Simplified the code to hide the check boxes. Here are the screenshots:
Calendar View:
image
List View:
image

frontend/src/pages/CalendarPage/components/Calendar/index.js
```diff
@@ -10,9 +10,29 @@ import {
...
 import { ViewState } from '@devexpress/dx-react-scheduler';
 
+// color of events based on status
+const resources = [
+  {
+    fieldName: 'status',
+    instances: [
+      { id: 'complete', color: '#b3cce6' },
+      { id: 'pending', color: '#e6b3cc' },
+      { id: 'ready', color: '#f28c8c' },
```
thaigillespie, line 24: I appreciate the change to hex colors!

frontend/src/pages/CalendarPage/components/Calendar/index.js
```diff
@@ -10,9 +10,29 @@ import {
...
+    ],
+  },
+];
+
+const mainResource = 'status';
+
+resources.propTypes = {
+  fieldName: PropTypes.string.isRequired,
+  instances: PropTypes.object.isRequired,
```
thaigillespie, line 33: Based on how you defined `instances` starting from line 21, the proptypes for it should be an array of objects, and not just an object

thaigillespie (changes requested):
Overall lgtm, just one small thing that needs changing

frontend/src/pages/CalendarPage/components/Calendar/index.js
thaigillespie: Well actually, since resources is not even a prop of the Calendar component, you don't even need these propTypes definitions. Can probably just remove them. One day most of our files will be changed to TypeScript and we'll have a better time with types : )

godwinpang (approved):
lgtm - @thaigillespie waiting on ur review then @yuw006 feel free to squash and merge :)

thaigillespie (approved):
Lgtm @yuw006 merge whenever you're free :)

PR #209, opened 2021-04-05
Hotfix for failing release phases due to migrations without SSL. 

godwinpang:
Release phases were previously run with NODE_ENV set to development and was connecting to prod db with SSL off. This PR adds `migration:run:ci` that runs without overriding NODE_ENV.


thaigillespie (approved):
Lgtm. Any reason why you used "ci" in the new command?

godwinpang:
I was thinking this would run on CI jobs which I now realize isn't true - would release be better?

thaigillespie:
Yeah I think that would be better, or just something like ":sslon" would work as well

thaigillespie (approved):

PR #210, opened 2021-04-05
Add path alias support in storybook.

godwinpang:
Storybook previously didn't support importing via path aliases - TsconfigPathsPlugin allows us to do so 💯


frontend/.storybook/main.js
```diff
@@ -11,4 +13,11 @@ module.exports = {
       },
     },
   ],
+  webpackFinal: async config => {
+    [].push.apply(config.resolve.plugins, [
```
thaigillespie, line 17: Just curious as to why you needed to use apply() here?

godwinpang: uh i guess this could be a map as well, honestly i don't know since this is copypasta - @thaigillespie do you think we need to change anything? this is just a config in .storybook

thaigillespie: Nah no need to change anything, I'm just a bit confused by that line

thaigillespie (approved):
Lgtm

PR #211, opened 2021-04-05
Multiple buckets support

ericke8:
Changed some variables around so that we can support multiple buckets in the environment


godwinpang (commented):
tysm for the very detailed documentation! 
everything looks good, but I just wanted to comment on some things to spark some discussion before we merge this in. 
1. Any future plans for handling more than a handful of buckets? We probably don't need to tackle this ever; just a nice to have.
2. As of now, we have config as essentially a global variable that's exported everywhere - that's fine until we want to write some e2e tests in the future that would hit the actual S3 buckets; is there any benefit to injecting config in the constructor? we could also use a separate s3 accessKey/secretKey, but i'm not sure if that requires another account

ericke8:
Part 1: If we get a lot of buckets, there isn't a really good way I can think of right now.  Since we are directly making requests to buckets on our side, we need to know each of the bucket names.  I'm not sure how we could have all of those without some kind of dictionary in config or env.
One (overcomplicated) way I can think of is simply leaving all the bucket management to an API or Lambda on AWS side, but that would require large changes to our design and may be more complicated/expensive, so I don't think it's really a feasible solution right now.

I have not thought too much on the problem of too many S3 buckets before, so I'm open to any discussion.


Part 2: Currently, the AWS stuff is loaded when we first run the StorageService file, using the creds in env.  If we want to do E2E tests with real buckets, we need to make another IAM user and make new buckets, so the test env would use a different key pair.  By that point we may as well change the bucket names in the env as well, so the test env file would be completely different.

That said, the config is currently loaded just as a part of the file, so it can easily just be moved into the class constructor if that makes it clearer.

godwinpang:
ye - sounds good @ericke8 :) we can probably talk a bit about the bucket stuff at the meeting

godwinpang (approved):

godwinpang:
before we merge could we get a +1 from @AimeeShao @Divneet18 @yuw006 or @Titan-Ngo ? just want to address any gaps in the docs since i'm probably not the best target audience

thaigillespie:
> 
> 
> before we merge could we get a +1 from @AimeeShao @Divneet18 @yuw006 or @Titan-Ngo ? just want to address any gaps in the docs since i'm probably not the best target audience

Adding Aimee, Divneet, Yueqi and Titan to list of reviewers per Godwin's comments

Titan-Ngo:
My comments about the documentation:
- overall, makes a lot of sense! 
- I was confused about the difference between a Service and a Controller but the GraphicService vs GraphicController example clears it all up.
- It also works well to clarify many of the changes you made (like adding an options parameter to downloadFile, adding ResumeBucket to the config file)
- Correct me if I'm wrong, but from my understanding: we have a bucket that holds a bunch of resumes (called ResumeBucket), and a ResumeService that actually does the uploading and downloading. Not sure if/where we have a ResumeController?

ericke8:
> * Correct me if I'm wrong, but from my understanding: we have a bucket that holds a bunch of resumes (called ResumeBucket), and a ResumeService that actually does the uploading and downloading. Not sure if/where we have a ResumeController?

Instead of a dedicated `ResumeController`, we simply call the `ResumeService` from the `UserController` for each individual user.

yuw006 (approved):
Looks good. Is there a place where we have a list of all current services?

PR #212, opened 2021-04-07
Pages refactor for readability.

godwinpang:
adding pages was getting sucky so I made this - idk if it's actually more readable so please comment


thaigillespie (approved):
Can probably have a separate file dedicated to storing these key-value pairs (of course this is not necessary). Otherwise I think this is great.

I'm sure you checked if the portal works after these changes right?

PR #217, opened 2021-04-10
Fix react-dev-utils due to conflict with storybook and pin material-t…

godwinpang:
…able due to jspdf problems in new version.

For me we have a devDependency of @storybook/react@^5.3.19 which has a dependency of react-dev-utils@9.1.0.
For some reason with this devDependency NODE_ENV=production yarn install put react-dev-utils in node_modules/react-scripts/node_modules/react-dev-utils instead of in node_modules/react-dev-utils.

If I uninstall @storybook/react (again, a devDependency...) NODE_ENV=production yarn install ; yarn build works fine.

Simplest workaround is explicitly install react-dev-utils.


thaigillespie (approved):
lgtm

PR #220, opened 2021-04-21
Update the event creation time to the nearest half hour

Divneet18:
This PR fixes the issue below 
https://github.com/HKN-UCSD/hkn-server/issues/187


godwinpang:
closes #187

frontend/src/pages/EventCreationPage/components/EventCreationForm/index.tsx
```diff
@@ -29,9 +29,29 @@ interface InitialValuesType {
   hosts: OfficerNameData[];
 }
 
+function findNearestHour(date, isStartDate) {
```
thaigillespie, line 32: Since this is just a helper function, I think it's better to use arrow function syntax here instead. Link: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

Also, since we are using Typescript (at least for this file), please add type annotations if possible to your function definition (https://www.typescriptlang.org/docs/handbook/2/everyday-types.html)

frontend/src/pages/EventCreationPage/components/EventCreationForm/index.tsx
```diff
@@ -29,9 +29,29 @@ interface InitialValuesType {
...
+  let hour = date.getHours();
+  let startMinutes = date.getMinutes();
+  if (startMinutes === 0) {
+    startMinutes = '00';
+  } else if (startMinutes <= 30) {
+    startMinutes = '30';
+  } else if (startMinutes > 30) {
+    startMinutes = '00';
+    hour += 1;
```
thaigillespie, line 41: It is much better to use **date-fns** (the datetime library we are using) to handle datetime arithmetic like what you're doing here, since it does all the overflow handling for you. Reference: https://date-fns.org/docs/Getting-Started

Specific function to add some amount of time to a specific date: https://date-fns.org/v2.21.1/docs/add

frontend/src/pages/EventCreationPage/components/EventCreationForm/index.tsx
```diff
@@ -29,9 +29,29 @@ interface InitialValuesType {
...
+  }
+
+  date.setMinutes(startMinutes);
```
thaigillespie, line 44: Please use [setMinutes by date-fns](https://date-fns.org/v2.21.1/docs/setMinutes) instead. I think you're using JS's Date API here? In the future, try to use date-fns as much as you can (you'll most likely never need to explicitly use Date API, but you might need to check its docs once in a while).

frontend/src/pages/EventCreationPage/components/EventCreationForm/index.tsx
```diff
@@ -29,9 +29,29 @@ interface InitialValuesType {
   hosts: OfficerNameData[];
 }
 
+function findNearestHour(date, isStartDate) {
+  let hour = date.getHours();
+  let startMinutes = date.getMinutes();
+  if (startMinutes === 0) {
+    startMinutes = '00';
```
thaigillespie, line 36: I think you should use a Number here instead of a string. I think for Date.setMinutes() and date-fns's setMinutes() they both have the minute param as type Number

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMinutes
https://date-fns.org/v2.21.1/docs/setMinutes

thaigillespie (changes requested):
The high level idea here is correct for snapping time to the nearest half hour. However, I think you should utilize date-fns a lot more here (though I'm guessing you probably didn't know it existed). date-fns is very useful and abstracts all the annoying details of the JS Date API away for you, so you should definitely use it.

Also some issues with TS syntax. Overall great first steps though!

godwinpang:
@thaigillespie could you also explain a little more about `date-fns?`

thaigillespie:
> 
> 
> @thaigillespie could you also explain a little more about `date-fns?`

Like explain what it is and what it does?

frontend/src/pages/EventCreationPage/components/EventCreationForm/index.tsx
```diff
@@ -29,9 +29,29 @@ interface InitialValuesType {
...
+  } else if (startMinutes <= 30) {
+    startMinutes = '30';
+  } else if (startMinutes > 30) {
+    startMinutes = '00';
+    hour += 1;
+  }
+
+  date.setMinutes(startMinutes);
+  date.setHours(hour);
+  if (!isStartDate) {
```
godwinpang, line 46: since the second `isStartDate` parameter only adds an hour to the endDate, `findNearestHour` would be more flexible if we removed the isStartDate parameter and instead did the hour adding outside the function since that preserves `findNearestHour` as doing that thing only, and allows us to reuse it elsewhere outside of the event start end date contexts.

see https://en.wikipedia.org/wiki/Inversion_of_control if you're interested in more details! :)

godwinpang:
> > @thaigillespie could you also explain a little more about `date-fns?`
> 
> Like explain what it is and what it does?

do we have links to a doc where we choose to use date-fns?

thaigillespie:
> 
> 
> > > @thaigillespie could you also explain a little more about `date-fns?`
> > 
> > 
> > Like explain what it is and what it does?
> 
> do we have links to a doc where we choose to use date-fns?

[Here](https://github.com/HKN-UCSD/hkn-server/blob/master/frontend/guides/miscellaneous/date_time_library_to_use.md) it is. @Divneet18 read through this if you can : )

Divneet18:
@thaigillespie @godwinpang I have made the requested changes. Please take a look.

frontend/src/pages/EventCreationPage/components/EventCreationForm/index.tsx
```diff
@@ -29,9 +29,29 @@ interface InitialValuesType {
   hosts: OfficerNameData[];
 }
 
+let findNearestHalfHour = (date: Date) => {
+  let startMinutes = getMinutes(date);
+  if (startMinutes === 0) {
+    startMinutes = 0;
```
thaigillespie, line 35: Don't think you need to set startMinutes to 0 if it is already 0

frontend/src/pages/EventCreationPage/components/EventCreationForm/index.tsx
```diff
@@ -29,9 +29,29 @@ interface InitialValuesType {
...
+  }
+  else if (startMinutes != 0 && startMinutes <= 30) {
+    date = add(date, { minutes: (30 - startMinutes) });
```
thaigillespie, line 38: This is fine, but since you're just setting it to 30 minutes anyway, this function is probably even easier to use: https://date-fns.org/v2.21.1/docs/setMinutes

thaigillespie (changes requested):
Good job on the changes. I've added my feedback for the changes you made. You can change them now if you want, but I reached out to Events and learned that we'll need to change the issue, since this is not exactly what they are looking for (snapping to nearest half hour on initial values of start date and end date only). I'll update the GitHub issue associated with your PR later to reflect updates as to what Events wants with this snapping hour thing. So yeah you can make changes based on my feedback now or just wait until I update the issue and we touch bases on this update to work on the issue : ) @Divneet18

thaigillespie:
Closing bc still unclear on what to do for the creation time stuff. If we work on this again, might as well make a new PR

PR #221, opened 2021-04-22
reformated calendar

### LINKED ISSUE #189: Calendar Overflow Horizontally
Calendar currently overflows if the window is not wide enough

yuw006:
Added a class in styles for the calendar component, which limits width to 100%
Week view and month view will not overflow:
overflow
fix #189


thaigillespie (approved):
Lgtm

PR #222, opened 2021-04-22
Adds pages for Forbidden and NotFound.

godwinpang:
This allows us to add handlers for 403 and 404 error codes later on in our network layer.


thaigillespie (approved):
lgtm

PR #248, opened 2021-07-12
Add more docs for FE and BE to help devs, and add docs for git

thaigillespie:


backend/guides/backend_api/hierarchical_connection.md
```diff
@@ -0,0 +1,37 @@
+# Hierarchical Connection Between Different Layers Of The Backend:
+
+## Current Layout:
+
+```
+             |------------ Mappers --------------------|
```
kyle1373, line 6: Would it be better to make the Controllers point to Mappers, Services, and Middlewares? On line 28, it mentions that Controllers depend on Mappers, Services, and Middlewares, yet Controllers isn't pointing to either of them.

If so, would the diagram look like this instead?

```
             |-----------> Mappers --------------------|
             |                 |                       |
Controllers--|                \|/                      |
             |-----------> Services ------------ Repositories ------------ Database
             |                /|\
             |                 |
             |-----------> Middlewares
```

thaigillespie: I specified on line 25 of the file that I only use the arrows for layers that are on the same level, just so readers know how layers on the same level depend on each other

kyle1373: I see. Makes sense now, thank you!

Titan-Ngo (commented):
Thanks for the docs Thai. Overall, everything is pretty detailed and clear. Maybe in the PR section, talk about the "closed" keyword since we weren't aware of it earlier?

thaigillespie:
> 
> 
> Thanks for the docs Thai. Overall, everything is pretty detailed and clear. Maybe in the PR section, talk about the "closed" keyword since we weren't aware of it earlier?

Can you pinpoint as to where that is? I'm not sure if I see it in the PR section. Maybe a good way to do this is when you conduct your review, make a comment on a relevant line of code in the file you want to make a comment on, that way I know more clearly what you're referring to

guides/git/pull_request_process.md
```diff
@@ -0,0 +1,16 @@
+# Process For Working With Pull Requests:
+
+## Overall Process In Several Steps:
+1. If you have just made a commit, GitHub will show an option to make a Pull Request (PR) for the branch you just committed to, near the top of your screen when you are on our repo's main GitHub page. Otherwise, go to the branch you want to make a PR for, click on `Contribute`, then click on `Open Pull Request` to open a PR for your branch.
+2. Add to the description of the PR if need be to properly describe what you did for your PR. This is especially important for frontend tasks, where you can add images and even videos of your progress/functionality.
+3. Assign yourself, as well as others who have worked on the same PR, under `Assignees` (or just remind others who have worked on the PR to add their names).
```
Titan-Ngo, line 6: Hi Thai, right here! It'll be great if you can add something about how we need to say "closes issue ___" as part of making a PR. just because a few of us (me, yueqi and more) were unaware of it before

Titan-Ngo (changes requested):

guides/git/pull_request_process.md
thai-tgil: Oh I don't think that was ever required of any of our team members. Not that it isn't a good practice though. I'll add some words regarding this in the docs

Titan-Ngo (approved):

thai-tgil:
@kyle1373 @yuw006 @audreygao can yall give a review as well?

kyle1373 (approved):
LGTM! Very helpful docs. Thanks for writing them!

PR #249, opened 2021-07-14
added snackbar to display information

yuw006:
This is intended for api errors, but to prove that it works I added it to toggle view in the calendar page
snackbar example
when view is switched this would display.
If anyone were to use it, they need to pass in open (toggled between on/off), message, severity ("warning","success","info","error"), handle close to close the snackbar. It belongs under components/errors. 
for more information go to 
https://material-ui.com/components/snackbars/


frontend/src/pages/CalendarPage/index.js
```diff
@@ -205,6 +218,14 @@ class CalendarPage extends React.Component {
                 )}
               </Paper>
             </Grid>
+<<<<<<< HEAD
```
thaigillespie, line 221: There's a merge conflict here, can you fix that? It's breaking your deploy previews

frontend/src/components/error/snackBar.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
thaigillespie, line 1: I think it would be better to make the component folder "snackbar" and not "error". Then, in src/components/snackbar, have a base snackbar (BaseSnackbar), then you can have things like ErrorSnackbar, InfoSnackbar, etc. that derive from it. You can use dropdowns and modals as examples for this (I recommend dropdowns, the PR to refactor modals is still stuck in hkn-member-portal so it's a bit hard to read through modals here in hkn-server).

If we do it this way, devs can just use ErrorSnackbar to display errors, InfoSnackbar to display info and not have to worry about customizing their snackbar. The point is that with these abstractions, we want to limit customization where we can, in order to have some sort of standardization in terms of these components (i.e. we don't want different ErrorSnackbar to look and be displayed differently from each other)

Let me know what you think about this.

frontend/src/pages/CalendarPage/index.js
```diff
@@ -215,6 +228,11 @@ class CalendarPage extends React.Component {
               </Container>
             )}
           </Grid>
+          <SnackbarErrors
+            open={openSnackbar}
+            message={message}
+            severity={severity}
+            handleClose={this.closeSnackbar()} />
```
thaigillespie, line 235: This is what's causing the infinite setState() loop error, causing your CircleCI test to fail. If you change it to `() => this.closeSnackbar()` it will work. handleClose needs a function, but what you're passing in is the function call itself and not a function

thaigillespie (changes requested):

thaigillespie:
Looks pretty good so far. Left some comments for you to look at!

frontend/src/components/error/snackBar.js
yuw006: Yah that sounds reasonable I'll rename them to something more appropriate

frontend/src/pages/CalendarPage/index.js
yuw006: oh thanks I was wondering what was up

frontend/src/components/alert/snackBar.tsx
```diff
@@ -0,0 +1,25 @@
...
+import Snackbar from '@material-ui/core/Snackbar';
+import MuiAlert from '@material-ui/lab/Alert';
+import Fade from '@material-ui/core/Fade';
+
+function Alert(props) {
+  return <MuiAlert elevation={6} variant="filled" {...props} />;
+}
+
+const SnackbarAlerts = props => {
+  const { open, message, severity, handleClose } = props;
```
thaigillespie, line 11: You'll need to make a TS interface and make `props` of that type to avoid the PropTypes error that eslint is giving.

TS interface: https://www.typescriptlang.org/docs/handbook/2/objects.html#interfaces-vs-intersections
Link to deprecated doc of TS interface but might still be useful: https://www.typescriptlang.org/docs/handbook/interfaces.html

thaigillespie (changes requested):
Just one thing I've noticed so far that needs changing so CI/CD for frontend will pass

frontend/src/components/alert/snackBar.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import Snackbar from '@material-ui/core/Snackbar';
+import MuiAlert from '@material-ui/lab/Alert';
+import Fade from '@material-ui/core/Fade';
+
+interface snackbarProps {
+  open: boolean;
+  message: string;
+  severity: string;
+  handleClose: undefined;
```
thaigillespie, line 10: Use `() => void` instead of undefined. Example here: https://github.com/HKN-UCSD/hkn-member-portal/blob/modal_refactor_new/src/components/modals/BaseModal.tsx

thaigillespie (approved):
Mostly good to me, just this one thing left. I'll approve after you fix it

thaigillespie (approved):
lgtm

thaigillespie (approved):
lgtm, since there is a merge conflict with backend/package-lock.json, I think it's best that you remove your backend/package-lock.json from your branch, then pull from master, then push these changes to your branch, and squash merge

Titan-Ngo:
Looks good to me, also waiting for the removal of the package-lock.json file before approving

Titan-Ngo:
Actually, since Thai approved it already, I'll approve it as well, just remember Thai's suggestion!

Titan-Ngo (dismissed):

frontend/src/pages/CalendarPage/index.js
godwinpang: can also just be `this.closeSnackbar`

thaigillespie (approved):
lgtm

PR #250, opened 2021-07-16
Add delete attendance endpoint

thaigillespie:
Add AttendanceController and endpoint to delete attendances, add skeleton code for updating attendances' times, add delete button on frontend to attendance table on event details page, will add that button
for check off table and, in another PR, RSVP table

fsdfdf

IGNORE THE EDIT BUTTON THAT IS JUST THERE FOR ME TO SEE IF BOTH CAN FIT INTO THE ATTENDANCE TABLE


yuw006:
Do you think that we need maybe a "confirm deletion" modal in case of a misclick? or would that be a bit annoying to the user

thaigillespie:
> 
> 
> Do you think that we need maybe a "confirm deletion" modal in case of a misclick? or would that be a bit annoying to the user

That's a good idea. We have that for event deletion too so I might as well just add it for this

thaigillespie:
Added confirmation modal

yuw006:
looks good to me

yuw006 (approved):

PR #253, opened 2021-07-27
Edit attendance UI and backend

thaigillespie:
Screenshot 2021-07-27 at 16-02-08 Member Portal Eta Kappa Nu Kappa Psi - HKN UCSD


Titan-Ngo (changes requested):
Beautiful code, is it okay for you to add some comments?

thaigillespie:
> 
> 
> Beautiful code, is it okay for you to add some comments?

Yep that's definitely fine. Where would you want some comments in the code that I've written?

thai-tgil:
> 
> 
> > Beautiful code, is it okay for you to add some comments?
> 
> Yep that's definitely fine. Where would you want some comments in the code that I've written?

@Titan-Ngo bump

backend/src/services/AttendanceService.ts
```diff
@@ -123,15 +123,28 @@ export class AttendanceService {
...
   }
 
-  async deleteAttendance(attendeeId: number, eventId: number): Promise<Attendance | undefined> {
+  async deleteAttendance(attendanceQuery: GetAttendanceQuery): Promise<Attendance | undefined> {
     const attendanceRepository = getRepository(Attendance);
+    const { attendeeId, eventId } = attendanceQuery;
+
     const attendance = await this.getAttendance(attendeeId, eventId);
 
     return attendance ? attendanceRepository.remove(attendance) : undefined;
```
Titan-Ngo, line 150: maybe a comment saying how/when attendanceRepo returns True and how/when it returns False causing undefined?

backend/src/services/AttendanceService.ts
```diff
@@ -123,15 +123,28 @@ export class AttendanceService {
     return points;
   }
 
-  async saveAttendance(attendance: Attendance): Promise<Attendance | undefined> {
+  async saveAttendance(
+    attendanceQuery: GetAttendanceQuery,
+    attendanceReq: AttendanceRequest,
+    officerId: number
+  ): Promise<Attendance> {
     const attendanceRepository = getRepository(Attendance);
```
Titan-Ngo, line 131: This might just be a 'me' thing, but maybe a reminder of what getRepository does and why we pass in Attendance?

frontend/src/services/AttendanceService.ts
```diff
@@ -1,9 +1,28 @@
...
+  AttendanceApi,
+  AttendanceControllerDeleteAttendanceRequest,
+  AttendanceControllerUpdateAttendanceTimesRequest
+} from './api/apis/AttendanceApi';
 import { AttendanceResponse } from './api/models';
 
 import ApiConfigStore from './ApiConfigStore';
 import { Configuration } from './api/runtime';
 
+export async function updateAttendanceTimes(attendeeId: number, eventId: number, startTime: string, endTime: string) {
```
Titan-Ngo, line 11: A comment about how updating startTime and endTime will also update attendee's induction points?  (which are automatically calculated from startTime and endTime)

Titan-Ngo (changes requested):

backend/src/services/AttendanceService.ts
thaigillespie: I think this is more appropriate as an information to be added in a documentation, which is already done [here](https://github.com/HKN-UCSD/hkn-server/blob/master/backend/guides/add/add_new_service.md). We use getRepository() in an ad hoc manner in a lot of places so it's kinda hard to put a comment for all of them

thaigillespie:
@Titan-Ngo added comments

thaigillespie:
Merge with master early to get the changes from master that gets rid of CircleCI cache so it doesn't mess up frontend CircleCI jobs

Titan-Ngo (approved):

Titan-Ngo (approved):

PR #266, opened 2021-08-18
Networking abstraction to better handle (a lot but not all) API calls

thaigillespie:
Created a useRequest hook that can be used with API calls in replacement of useState + useEffect. Also made two components to handle errors resulting from error response from the API calls caught by the useRequest hook -> RequestErrorSnackbar and RequestErrorModal.

Scenario for screenshots below: Accessing an EventDetailsPage of a non-existing event. Before, since we left the error response unhandled, it didn't have any indication that the GET operation failed. On the production portal, it would just stay there, with the loading bar going on forever. Even worse, in locally run portal, one would get an error screen when the GET operation fails. This resolves the issue detailed above pretty nicely, and gives clear and visual indication of what happened.

RequestErrorSnackbar:
requestErrorSnackbar

RequestErrorModal:
requestErrorModal


frontend/src/components/alert/snackBar.tsx
```diff
@@ -8,19 +8,22 @@ interface snackbarProps {
   message: string;
   severity: string;
   handleClose: () => void;
+  autoHideDuration?: number | null;
```
yuw006, line 11: so what happens when it's null?

frontend/src/components/request/index.ts
```diff
@@ -0,0 +1,4 @@
+import RequestErrorModal from './RequestErrorModal';
+import RequestErrorSnackbar from './RequestErrorSnackbar';
+
+export { RequestErrorModal, RequestErrorSnackbar };
```
yuw006, line 4: do you think it would be a good idea to export snackBarAlerts here as well?

yuw006 (commented):
Everything looks good, I just had two questions and I'll approve it.

frontend/src/components/alert/snackBar.tsx
thaigillespie: https://material-ui.com/api/snackbar/ if that is null then snackbar doesn't auto hide

frontend/src/components/request/index.ts
thaigillespie: I think we don't need to. My reasoning as to why I made a separate modal and snackbar for components/request is that since request components are only meant to be used for API calls, so I didn't put RequestErrorSnackbar in the general snackbar folder and same for RequestErrorModal as well. I just leave them here bc they're really specific to request and is not supposed to be used outside of rendering modal/snackbar based on whether an error occurred for a request or not. Since snackbarAlerts can be used for a lot of purposes, it is better imo to leave it where it is.

yuw006: Sounds good

yuw006 (approved):
lgtm

frontend/src/pages/App/index.tsx
```diff
@@ -98,85 +101,88 @@ function App(): JSX.Element {
...
+          <Switch>
+            <Route
+              exact
+              path={ROUTES.SIGN_IN}
+              render={() => <SignInPage setClaims={setClaims} />}
+            />
+            <Route exact path={ROUTES.SIGN_UP} render={() => <SignUpPage />} />
+            <Route
+              exact
+              path={ROUTES.EVENT_SIGN_IN}
```
Titan-Ngo, line 116: I just merged the qr code branch into master, so you can update the routes to reflect that when merging :)

Titan-Ngo (approved):
lgtm

frontend/src/pages/App/index.tsx
thaigillespie: Yeah for sure, thanks!

PR #267, opened 2021-08-24
Connect developed components together for first major task

thai-truong:
Flow:

1. Start at qrcode-intermediate page for event with id = 25 and user is not signed in (can't start at qrcode page bc we needed the full URL with protocol and domain for the QR code, but this was hardcoded so we don't have access to that in the code - this is fine, at least for now):

qrCodeIntermediateStart

2. Since user is not signed in, qrcode-intermediate redirects to eventSignInOptions page:

qrCodeIntermediateNotSignedInLeadsToEventSignInOptions

3. Guest event sign in page pops up when clicked on "Guest Event Sign In" on eventSignInOptions page:

EventSignInOptionsGuestEventSignIn

4. Normal account sign in page pops up when clicked on "Inductee / Member Log in" on eventSignInOptions page:

EventSignInOptionsInducteeMemberSignIn

5. Redirects straight to the event with id = 25 after signing in to account:

GoesStraightToCorrectEventDetailsPageAfterSigningIn

6. Start again at qrcode-intermediate page, this time user is signed in and the event is also event with id = 25:

qrCodeIntermediateStartWhenSignedIn

7. Redirects straight to the event with id = 25 because user is already signed in:

GoestStraightToCorrectEventDetailsFromLoggedInQRCodeIntermediate


thai-truong:
@Titan-Ngo @kyle1373 @yuw006 @audreygao  Please review this PR soon, it's been sitting here for a while now and there are only 5 files to review. We also need this in to be done with the first major task so please make comments if needed so I can get the changes in before we merge this to master

Titan-Ngo (approved):
lgtm, thanks for removing what I forgot to remove

PR #274, opened 2021-09-19
Change event signin page to fit in phone

audreygao:
How it looks on browser:
Screen Shot 2021-09-18 at 5 23 38 PM
Screen Shot 2021-09-18 at 5 24 07 PM
On Phone:
IMG_5292
Screen Shot 2021-09-18 at 5 24 54 PM


frontend/src/pages/EventSignInPage/styles.ts
```diff
@@ -12,6 +12,16 @@ const useStyles = makeStyles({
     alignItems: 'center',
     width: '600px',
     paddingBottom: '30px',
+    [theme.breakpoints.up('sm')]: {
+      width: 600,
+      marginLeft: 'auto',
+      marginRight: 'auto',
+    },
+    [theme.breakpoints.down('sm')]: {
```
thai-truong, line 20: Just out of curiosity, what does the up() and down() thing do? And what is the `breakpoints` thing? I've always been bad with CSS so just wanted to know : )

frontend/src/pages/EventSignInOptionsPage/index.tsx
```diff
@@ -21,7 +21,7 @@ function EventSignInOptionsPage(): JSX.Element {
   const guestSignInURL = ROUTES.EVENT_SIGN_IN_WITH_ID(eventId);
 
   return (
-    <div style={{ margin: '200px' }}>
```
thai-truong, line 24: Nice catch!

thai-truong (approved):
Lgtm! Thanks for doing this, it looks so much nicer. I'm not very good at CSS so I say you should wait to the other reviewer to give an approval before merging, but definitely an approval from me.

frontend/src/pages/EventSignInPage/styles.ts
audreygao: it's creating different size media query based on the screen size, up would be targeting screen size larger than the given parameter, and down would be for screen size smaller than the parameter

thai-truong:
@yuw006 can you review this PR since it should be quick?

yuw006:
Just curious what happens when the details is really long?

audreygao:
> Just curious what happens when the details is really long? @audreygao

like which part? It would just wrap the text I believe.

yuw006:
> > Just curious what happens when the details is really long? @audreygao
> 
> like which part? It would just wrap the text I believe.

sounds good

yuw006 (approved):

thai-truong:
You can merge this in now @audreygao

PR #277, opened 2021-09-21
change backend process for signup

### LINKED ISSUE #256: Revamp account sign up process in the backend
Right now, our account sign up process (for inductees) is:
- Fill out HKN inductee form to become an inductee -> Store all the info on Google Sheets -> Get the CSV file for all that info -> Run a Python script to create AppUser entities for all the ppl included in that CSV file -> Ppl with an existing initial AppUser entity created from the Python script sign up for an account using the sign up page on the portal -> An account is created and they can use their credentials to access the portal.

What we want to change to is:
- Fill out sign up form on the account sign up page on the portal -> New account is created for them to use after they fill out everything and submit the form as a new user.

To implement this new flow, you need to change the implementation of `signUpUser()` handler in `AuthController` in the backend. Right now, based on the current flow, `signUpUser()` finds an existing AppUser (created from CSV file via Python script) based on the submitted email, then creates a new Firebase account for them, then updates that existing AppUser entity with the new information submitted through the form. With our new flow, things will be much simpler. You just need to somehow create a new AppUser entity based on the info passed in from the form and create a new Firebase account, and return this new AppUser entity. Remember to add a check to reject the sign up request if they already have an account.

For this task, you most likely have to change the `AppUserSignupRequest` payload to properly reflect the changes we are making to the information we are gathering in the sign up form. This means that you'll have to run autogen on frontend so frontend has the latest updates to API calls to do account sign up. Then, you'll have to modify `createNewUser()` in frontend's `AuthService.ts` to accommodate changes to the account sign up process.

yuw006:
close #256 
picture of post request for the same account twice
results


backend/src/mappers/AppUserMapper.ts
```diff
@@ -18,15 +19,27 @@ export class AppUserMapper {
...
    * @returns {AppUser} A newly created AppUser entity.
    */
   async requestToNewEntity(
-    appUserRequest: AppUserPostRequest | AppUserEventRequest
+    appUserRequest: AppUserPostRequest | AppUserEventRequest | InducteePostRequest
   ): Promise<AppUser> {
     const appUserRepository = getRepository(AppUser);
     const inductionClassRepository = getRepository(InductionClass);
 
+    if (appUserRequest instanceof InducteePostRequest) {
```
thai-truong, line 31: Since this if block contains the same content as the one below it, you can just combine the conditionals i.e. `appUserRequest instanceof InducteePostRequest || appUserRequest instanceof AppUserPostRequest` or something like that

backend/src/payloads/AppUser.ts
```diff
@@ -261,3 +261,49 @@ export class AppUserInterviewAvailability {
...
+  readonly lastName: string;
+
+  @IsString()
+  readonly major: string;
+
+  @IsString()
+  readonly graduationYear: string;
+
+  @IsString()
+  readonly password: string;
```
thai-truong, line 282: I don't see you using password for inducteePostRequest so we can remove this I think.

backend/src/payloads/AppUser.ts
```diff
@@ -261,3 +261,49 @@ export class AppUserInterviewAvailability {
...
+
+  @IsString()
+  readonly preferredName: string;
```
thai-truong, line 285: `preferredName`, `pronoun` and `customPronoun` are optional on the new inductee account sign up page so they should be optional here as well. While you're at it, do the same thing here for AppUserSignupRequest

backend/src/payloads/AppUser.ts
```diff
@@ -261,3 +261,49 @@ export class AppUserInterviewAvailability {
   @IsDateString()
   end: string;
 }
+
+export class InducteePostRequest {
```
thai-truong, line 265: Should put this payload class with the Request payload classes, probably after AppUserSignupRequest, to group  them up nicely together

thai-truong (changes requested):
Looks pretty good so far! Left you some comments to look at

backend/src/payloads/AppUser.ts
thai-truong: Actually I take it back, don't touch AppUserSignupRequest. You can just change the payload class you made

thai-truong (dismissed):
lgtm

thai-truong (approved):
lgtm

PR #279, opened 2021-09-24
New signup workflow integration

thai-truong:
Integrates new signup workflow into portal


thai-truong:
Need not review the autogenerated files i.e. files whose file path start with `frontend/src/services/api`.

thai-truong:
@yuw006 @kyle1373 @Titan-Ngo @audreygao y'all only need to review the first 19 files out of all 30 files, aka files that are not the auto generated ones

backend/src/controllers/AuthController.ts
```diff
@@ -108,40 +136,20 @@ export class AuthController {
...
     );
 
-    if (newAccountID === undefined) {
-      return undefined;
-    }
-
+    const savedAppUser: AppUser = await this.appUserService.saveAppUser(appUserToSave);
     return this.appUserMapper.entityToResponse(savedAppUser);
-  }
+  } */
```
yuw006, line 147: why is it commented not deleted?

frontend/src/services/AuthService.ts
```diff
@@ -1,14 +1,24 @@
...
+  appUserSignupRequest: AppUserSignupRequest
+): Promise<AppUserResponse> {
+  const apiConfig: Configuration = await ApiConfigStore.getApiConfig();
+  const authApi: AuthApi = new AuthApi(apiConfig);
+  const request: AuthControllerInducteeSignUpRequest = { appUserSignupRequest };
+
+  return authApi.authControllerInducteeSignUp(request);
+}
+
+/*export async function createNewUser(
```
yuw006, line 16: also why is it commented?

backend/src/controllers/InductionClassController.ts
```diff
@@ -48,7 +48,7 @@ export class InductionClassController {
   @OpenAPI({ security: [{ TokenAuth: [] }] })
   async createInductionClass(
     @Body() inductionClassRequest: InductionClassRequest
-  ): Promise<InductionClassResponse> {
+  ): Promise<InductionClassResponse | undefined> {
```
yuw006, line 51: what situation is undefined for

backend/src/controllers/AuthController.ts
thai-truong: Usually I would delete unused code, but in this case I chose to keep them in commented form. This is because we might need the old sign up workflow in the future for whatever reason, maybe this new signup workflow doesn't work as well as we thought it would so we want to switch it back. After a while, when we're sure that we will stick with the new signup workflow, we can just remove the old workflow completely

frontend/src/services/AuthService.ts
thai-truong: Simply because the old signup workflow used this. Reasoning is the same as my response above.

backend/src/controllers/InductionClassController.ts
thai-truong: If you check `inductionClassService`'s createInductionClass(), you'll see that if the induction class passed in is a duplicate of an existing induction class, it will return undefined

yuw006 (approved):
everything looks good

PR #283, opened 2021-09-30
Updated bottom of sign-up form

timstjiang:
Adding a comment to contact me if they have an account already

### DIFF
--- a/frontend/src/pages/InducteeSignUpPage/components/SignUpForm/index.tsx
+++ b/frontend/src/pages/InducteeSignUpPage/components/SignUpForm/index.tsx
@@ -253,7 +253,7 @@ export const SignUpForm = (props: SignUpFormProps) => {
 			<Grid item>
               <Grid container justify='center'>
 				<Typography>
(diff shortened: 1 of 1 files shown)


frontend/src/pages/InducteeSignUpPage/components/SignUpForm/index.tsx
```diff
@@ -250,16 +250,18 @@ export const SignUpForm = (props: SignUpFormProps) => {
                 Sign Up
               </Button>
             </Grid>
-
-            <Grid item>
+			<Grid item>
               <Grid container justify='center'>
+				<Typography>
+				Have an account from previous cycles? Please contact Tim at s9jiang@ucsd.edu to update your existing account.
```
thai-truong, line 256: probably use induction cycles here instead of cycles for max clarity?

thai-truong (changes requested):
Just one thing. I'll approve after that is resolved

thai-truong (approved):
lgtm

PR #284, opened 2021-10-03
update new induction requirements in FE and BE

audreygao:


audreygao:
Screen Shot 2021-10-03 at 3 31 54 PM
Screen Shot 2021-10-03 at 3 32 39 PM

audreygao:
Added event type to attendance card and make 'complete' green and 'incomplete' red on inductee account points view
Screen Shot 2021-10-10 at 3 01 43 PM

frontend/src/pages/PointsPage/components/PointDisplay/index.tsx
```diff
@@ -29,6 +29,9 @@ export function PointDisplay(props: PointProps): JSX.Element {
                       <Typography variant='h5' component='h2'>
                         {event.name}
                       </Typography>
+                      <Typography color='textSecondary'>
+                        {event.type} event
```
thai-truong, line 33: I think it would be great if you can capitalize both `event.type` and the string `"event"` here. What do you think?

frontend/src/pages/PointsPage/components/PointDisplay/index.tsx
```diff
@@ -29,6 +29,9 @@ export function PointDisplay(props: PointProps): JSX.Element {
...
+                      </Typography>
                       <Typography className={classes.pos} color='textSecondary'>
                         {format(parseISO(attendance.startTime), 'PP')}
```
thai-truong, line 36: Idk if there is another suitable color for the attendance date but imo it's better to have the color be at least a bit different from the event type. Thoughts?

PR #291, opened 2021-10-18
edit event modal

yuw006:
https://user-images.githubusercontent.com/63177486/137767991-12247b0d-9d96-4323-996b-111198ecd002.mp4

The edited event modal would function as follows, and this would only apply to officers


frontend/src/pages/CalendarPage/index.js
```diff
@@ -101,6 +104,25 @@ class CalendarPage extends React.Component {
     });
   }
 
+  handleUpdateStatus = async (
```
thai-truong, line 107: You need not pass in the event object here. Notice what you're passing down to EventCard's event prop, you're passing down selectedEvent already. This means you really only need to pass in the updated status to this handleUpdateStatus() and create a modified selectedEvent from the original selectedEvent whose status is changed to the new status. Think carefully about what you're passing down to functions since you might not need them at all to do what you want.

frontend/src/pages/CalendarPage/index.js
```diff
@@ -101,6 +104,25 @@ class CalendarPage extends React.Component {
...
+        return {
+          id: host.id,
+        };
+      }),
+    };
+    await updateEvent(values.id, eventRequest);
+    const { eventStatusChanged, selectedEvent } = this.state;
+    this.setState({
+      eventStatusChanged: !eventStatusChanged,
+      selectedEvent: {...selectedEvent, status: newStatus}
```
thai-truong, line 122: Should use the result of updateEvent here. You can create a const var storing the result of updateEvent...

frontend/src/pages/CalendarPage/index.js
```diff
@@ -101,6 +104,25 @@ class CalendarPage extends React.Component {
     });
   }
 
+  handleUpdateStatus = async (
+    values, newStatus
```
thai-truong, line 108: Rename `values` to something else for more clarity since we clearly know an event object is being passed down. Tho you can probably just remove it altogether, per the comment above.

frontend/src/pages/CalendarPage/index.js
```diff
@@ -18,12 +18,14 @@ import * as ROUTES from '@Constants/routes';
...
 import { getAllEvents } from '@Services/EventService';
 import { Button } from '@SharedComponents';
+import { updateEvent } from '@Services/EventService';
 
 class CalendarPage extends React.Component {
   constructor() {
     super();
     this.state = {
       events: [],
+      eventStatusChanged: true,
```
thai-truong, line 28: hasEventStatusChanged is better imo

frontend/src/pages/CalendarPage/components/EventCard/index.js
```diff
@@ -30,6 +31,26 @@ function EventCard({ event, onClose, classes }) {
             <RoomIcon color='disabled' />
             <GetLocation location={event.location} />
           </Box>
+          <Grid>
+            {OfficerRenderPermission(Button)({
+              secondary: true,
+              disabled: event.status === 'pending',
+              children: 'Pending',
+              onClick: () => { const newEvent = { ...event, status: 'pending'}; updateStatus(newEvent, 'pending');},
```
thai-truong, line 39: Can reduce all of this to a simple updateStatus call, see comment in calendar page.

This applies for the other status buttons as well

frontend/src/pages/CalendarPage/index.js
yuw006: do you mean I should use it for selectedEvent or both?

PR #313, opened 2022-04-12
Add Name column to inductee points page

laurenkiyomi:
Screen Shot 2022-04-12 at 1 27 45 PM


frontend/src/services/PointsService.ts
```diff
@@ -36,6 +38,8 @@ export async function getAllInducteePoints(): Promise<InducteePoint[]> {
       hasSocialRequirement: point.hasSocialRequirement
         ? 'Complete'
         : 'Incomplete',
+      name: (await getUserById(point.user)).firstName + " " + (await getUserById(point.user)).lastName,
```
godwinpang, line 41: each getUserByID is an API call - we should try to just call it once inside the map func

laurenkiyomi: fixed

godwinpang (approved):
lgtm thx

audreygao (approved):

PR #325, opened 2022-05-24
add unrsvp button to event page

audreygao:


backend/src/controllers/EventController.ts
```diff
@@ -268,6 +272,38 @@ export class EventController {
 
     return this.rsvpMapper.entityToResponse(newRSVP);
   }
+
+  @Delete('/:eventID/unrsvp/affiliate')
```
godwinpang, line 276: if we are treating rsvp as an entity, then probably the following is a better endpoint name because we are DELETE-ing the rsvp
```suggestion
  @Delete('/:eventID/rsvp/affiliate')
```

backend/src/controllers/EventController.ts
```diff
@@ -268,6 +272,38 @@ export class EventController {
...
+  @OpenAPI({ security: [{ TokenAuth: [] }] })
+  async affiliateEventUnRSVP(
+    @Param('eventID') eventID: number,
+    @CurrentUser({ required: true }) appUser: AppUser
+  ): Promise<RSVPResponse | undefined> {
+    const getRSVPQuery = { eventId: eventID, appUserId: appUser.id };
+    const deletedRSVP = await this.rsvpService.deleteAffiliateRSVP(getRSVPQuery);
+
+    if (deletedRSVP === undefined) {
+      return undefined;
```
godwinpang, line 287: should we return an error instead?

audreygao: I see. I'll take a look and make changes after finals week.

PR #329, opened 2022-08-29
Add social media icons

erwok:


erwok:
Oops! I included 2 commits from before that don't work!

erwok:
I'm confused on why backend is failing. I haven't touched it.

erwok:
After ten years 🙏🙏🙏

backend/src/legacy/EventController.test.ts
```diff
@@ -11,10 +11,13 @@ beforeAll(async () => {
...
   testApp = app;
   testConnection = connection;
+  console.log(testApp, testConnection);
 });
 
 afterAll(async () => {
-  await testConnection.close();
+  if (testConnection != null) {
+    await testConnection.close();
+  }
```
kyle1373, line 20: For this file, it's OK to just not put any changes. Main already has this test under a legacy folder. I believe what should happen is when merging the branch, it will only merge changed files, so leaving this out is alright.

frontend/src/components/NavBar/styles.js
```diff
@@ -33,6 +33,33 @@ const styles = theme => ({
   h5: {
     marginBottom: theme.spacing(2),
   },
+  container: {
+    width: '90%',
+    display: 'flex',
+    flexDirection: 'column-reverse',
+    height: '100%',
+    borderCollapse: 'collapse',
+    paddingLeft: '2px',
```
kyle1373, line 42: I would make `paddingLeft` with the value `0px`. This would center the buttons instead of having them slightly more aligned to the right than normal

frontend/src/components/NavBar/styles.js
```diff
@@ -33,6 +33,33 @@ const styles = theme => ({
   h5: {
     marginBottom: theme.spacing(2),
   },
+  container: {
```
kyle1373, line 36: This shouldn't have the name `container` as it seems too general. Maybe a better name would be `socialContainer`.

kyle1373 (changes requested):
Great implementation! Please check the two comments I made, but beyond that, it looks good to me.

frontend/src/components/NavBar/styles.js
erwok: Added!

frontend/src/components/NavBar/styles.js
erwok: Added!

erwok (commented):
I incorporated the suggestions

kyle1373 (approved):

frontend/src/components/NavBar/socialButtons.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const LinkedInButton = {
+  pathname: "https://www.linkedin.com/company/hkn-ucsd/about/",
+  icon: <LinkedInIcon fontSize='inherit' />,
+  target: "_blank"
+}
+
+
+export const SocialButtons = [
```
Adneths, line 32: Since it's not actually exporting a Button element, it may be clearer to rename it to Info or Link instead of button, same for other instances and file name.

Adneths (approved):
Looks good overall

laurenkiyomi (dismissed):
Looks good to me other than renaming the "Button" elements.

PR #330, opened 2022-09-29
Edit roles page with admin privileges

kyle1373:
(TO BE FILLED IN LATER)


kyle1373:
WIll fix the tests in the future

backend/ormconfig.js
```diff
@@ -18,8 +18,8 @@ module.exports = [
   {
     type: 'postgres',
     url: process.env.DATABASE_URL,
-    ssl: sslConfig.ssl,
-    extra: sslConfig.extra,
+    ssl: prodSSLConfig.ssl,
```
godwinpang, line 21: why are we changing this?

frontend/src/components/NavBar/index.js
```diff
@@ -64,6 +70,17 @@ class NavBar extends React.Component {
...
 
+  handleTabs = () => {
+    switch (this.state.role) {
+      case 'admin':
+        return AdminTabs;
+      case 'officer':
+        return OfficerTabs;
+      default:
+        return InducteeTabs;
+    }
```
godwinpang, line 81:
```suggestion
  getTabsByRole = (role: string) => {
    switch (role) {
      case 'admin':
        return AdminTabs;
      case 'officer':
        return OfficerTabs;
      default:
        return InducteeTabs;
    }
```

easier to unit test

frontend/src/constants/routes.js
```diff
@@ -11,6 +11,7 @@ export const EVENT_EDIT = '/events/:eventId/edit';
 export const EVENT_DETAILS = '/events/:id';
 export const EVENT_CREATION = '/events?create=true';
 export const TEST = '/test';
+export const EDIT_ROLES = '/editroles';
```
godwinpang, line 14:
```suggestion
export const EDIT_ROLES = '/roles';
```
in general routes don't have verbs

frontend/src/pages/EditRolesPage/components/MemberTable/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+          email={email}
+          firstName={firstName}
+          lastName={lastName}
+          role={role}
+        />
+      ),
+    },
+  ];
+
+  const ComponentToRender = () => {
```
godwinpang, line 67: it's been a while, but can't we just return the function instead of naming it `ComponentToRender?`

frontend/src/pages/EditRolesPage/components/MemberTable/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+    name: firstName + ' ' + lastName,
+    email,
+    role,
+  };
+
+  return userRow;
+};
+
+function MemberTable(): JSX.Element {
+  const { data, isLoading, error, isError } = useRequest<
```
godwinpang, line 40: can we push all the `useRequest` calls up to the page level component, then pass in the responses down to the table as a prop? makes things easier to read

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
+import React, { Profiler } from 'react';
```
godwinpang, line 1: why do we need a Profiler import here?

frontend/src/pages/EditRolesPage/index.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Grid } from '@material-ui/core';
+
+import MemberTable from './components/MemberTable';
+
```
godwinpang, line 5: in general (well for this codebase), we tried to do all useRequests up at the page level

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+    return (
+      <div>
+        <Button positive primary onClick={this.handleEditRolePopup}>
+          Edit Role
+        </Button>
+        <Dialog
+          open={this.state.isPopupOpen!}
+          onClose={this.handleEditRoleClose}
+          aria-labelledby='alert-dialog-title'
+          aria-describedby='alert-dialog-description'
```
godwinpang, line 106: Oh we don't need these in this codebase, they're used by google for screenreaders but we don't support those anyways

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+    try {
+      await updateUserById(this.state.id!, updatedUser);
+      this.setState({
+        role: newRole,
+      });
+    } catch (error) {}
+    window.location.reload();
+  }
+
+  //const res: AppUserResponse = await getUserById(parseInt(userId, 10));
```
godwinpang, line 95: remove commented code

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+      email: this.state.email!,
+      firstName: this.state.firstName!,
+      lastName: this.state.lastName!,
+    };
+    try {
+      await updateUserById(this.state.id!, updatedUser);
+      this.setState({
+        role: newRole,
+      });
+    } catch (error) {}
```
godwinpang, line 91: nit: do we need to show a popup here for the error case

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+    window.location.reload();
```
godwinpang, line 92: we shouldn't be refreshing the window from a react component, what are we trying to do here?

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+interface INITIAL_STATES {
+  isPopupOpen?: boolean;
+  id?: number;
+  role?: string;
+  email?: string;
+  firstName?: string;
+  lastName?: string;
+}
+
+class EditRolesButton extends React.Component<
```
godwinpang, line 49: I think we can just make this a functional component? Do we use class components anywhere else in the codebase? (sorry it's been a while)

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+              color='primary'
+            >
+              Officer
+            </Button>
+            <Button
+              onClick={() => this.updateUserRole('admin')}
+              color='primary'
+            >
+              Administrator
+            </Button>
```
godwinpang, line 142: this is a small thing, but you can also do the following

```js
const roles = ['Admin', 'Officer',...];
return roles.map(role => {
    <Button 
        onClick={() => this.updateUserRole(role)}
        color='primary'
    >
        {role.toLowerCase()}
    </Button>
})
```

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+  handleEditRolePopup() {
+    this.setState({
+      isPopupOpen: true,
+    });
+  }
+  handleEditRoleClose() {
+    this.setState({
+      isPopupOpen: false,
+    });
+  }
```
godwinpang, line 78: maybe sth like the below? probably don't need 2 functions
```suggestion
  setShowPopup(shouldShowPopup bool) {
    this.setState({
      isPopupOpen: shouldShowPopup,
    });
  }
```

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { AppUserPostRequest } from '@Services/api';
+
+interface EditRolesButtonProps {
+  id: number;
+  role: string;
+  email: string;
+  firstName: string;
+  lastName: string;
+}
+interface INITIAL_STATES {
```
godwinpang, line 40: This `INITIAL_STATES` pattern is pretty old (i think pre functional components?); I remember writing stuff like this when I was first learning react 4 years ago

godwinpang (changes requested):

backend/ormconfig.js
kyle1373: This was originally there for me personally when connecting with the dev DB. Before I finish this PR I plan on going ahead and reverting this file back. I use npm instead of docker

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
kyle1373: This is unused and I plan on removing this upon cleaning up my changes

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
kyle1373: I was planning on adding an error popup sometime this week, but yeah great reminder 👍

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
kyle1373: This was a temporary fix for testing. I was trying to figure out a way to make it so the edited roles we submit upon popup show up on the table, since the rows themselves are only fetched once instead of updated automatically. I'm still trying to figure out a solution for this, so if you have any suggestions I'd greatly appreciate them. The rows aren't state variables either which makes it a bit harder for me to figure out.

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
kyle1373: We use a mix of class and functional components in the codebase I believe. In this case it's fine for it to be a functional component. I'll go ahead and make that change accordingly.

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
kyle1373: Would need to do some refactoring but yeah it does shorten it up a bit. I'll also implement this.

frontend/src/pages/EditRolesPage/components/MemberTable/index.tsx
kyle1373: In hindsight yeah this is a better change. I remember writing this when I was first learning react. Good catch

frontend/src/pages/EditRolesPage/components/MemberTable/index.tsx
kyle1373: I'll check this out more in detail and will follow up. I believe initially I had some issues with referencing the types from the MultipleAppUserResponse imports when just returning the function but worked when putting it under a component. Again, I'll look more deep into this.

frontend/src/pages/EditRolesPage/components/buttons/EditRolesButton.tsx
godwinpang: `EditRolesButton` can take in a callback from the parent component that sets the state of the parent component

PR #349, opened 2022-11-11
Feature/inductee signup alerts

### LINKED ISSUE #341: Receive feedback when an inductee already has an account on the inductee sign up page
Right now when we distribute invite links to inductees, they are able to create an account by clicking create account and getting a popup saying their account is created. However, there may be a chance an inductee already has an account in the HKN portal. When this happens, no popup actually occurs saying so, which can confuse the inductee thinking that creating an account doesn't actually work. We want to have a popup shown to the inductee which describes that their account has already been created and that they should log in with their same email address instead of just showing no popup at all.

This task is purely a frontend task and should be relatively straightforward. All that needs to be done is an extra check and a popup based on if that condition is met.

nnazeem:
To close #341


frontend/src/pages/InducteeSignUpPage/index.tsx
```diff
@@ -47,27 +47,25 @@ function InducteeSignUpPage(): JSX.Element {
...
 
     try {
       await doSendVerificationEmail();
     } catch {
-      console.log('Send verification email failed.');
+      alert('Send verification email failed.');
     }
     await doSignOut();
-    setSubmitting(false);
-    alert('You have successfully signed up for an account.');
```
kyle1373, line 70: Why are we taking out setSubmitting and the alert afterwards? Shouldn't they be useful to determine that account creation was successful?

kyle1373 (changes requested):
Everything looks good except for the concern with deleting line 69 and 70 on inducteesignuppage. Please take a look. Thanks!

frontend/src/pages/InducteeSignUpPage/index.tsx
Lord-of-Bugs: Isn't it taken out because the verification failed following the alert on line 66?

